PR #1, opened 2021-11-05
✨ run-analysis action files

laurentsimon:
Files to run the scorecard action.
See this file to see how to use it from a workflow https://github.com/laurentsimon/scorecard-action-test-4/blob/main/.github/workflows/scorecard-analysis.yml#L24
The branch is set to this PR branch for testing only - by default main branch is used and it not necessary.

### DIFF
--- a/run-analysis/entrypoint.sh
+++ b/run-analysis/entrypoint.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+# Copyright 2021 Security Scorecard Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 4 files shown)


run-analysis/entrypoint.sh
```diff
@@ -0,0 +1,51 @@
...
+fi
+
+# It's important to change directories here, to ensure
+# the files in SARIF start at the source of the repo.
+# This allows GitHub to highlight the file.
+cd "$GITHUB_WORKSPACE"
+
+if [[ "$GITHUB_EVENT_NAME" == "pull_request"* ]]
+then
+    $SCORECARD_BIN --local . --format sarif --show-details --policy="$SCORECARD_POLICY_FILE" > "$SCORECARD_SARIF_FILE"
```
asraa, line 46: just curious, will this depend on the repo being checked out? (i.e. use actions/checkout somewhere) -- maybe this is already done?

asraa (approved):
Thanks! LGTM

Assuming eventually you'd like a README in this repo (esp to define the policy, since there's no default one) -- happy to start a PR for this unless you have some copy-paste-able doc already ready!

run-analysis/entrypoint.sh
laurentsimon: absolutely https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml#L22

asraa: thanks!

asraa (approved):

naveensrinivasan:
## Scorecard Results</br>
</br>

Issue #5, opened 2021-11-09
Make structure compatible with GitHub marketplace


oliverchang:
Would it be possible to follow the required structure [here](https://docs.github.com/en/actions/creating-actions/publishing-actions-in-github-marketplace#about-publishing-actions) to make this more easily discoverable on the marketplace? 

The main requirement is that `action.yml` lives in the root (and that there is only 1 action per repo).

laurentsimon:
ah, was not aware of this requirement. We certainly can. It's a little inconvenient but I don't think there's a way around it. Thanks for pointing this out!

PR #8, opened 2021-12-02
Feature: move files to support market place publishing

laurentsimon:
The market place requirements are in https://docs.github.com/en/actions/creating-actions/publishing-actions-in-github-marketplace#about-publishing-actions

closes https://github.com/ossf/scorecard-action/issues/5


naveensrinivasan (approved):
Thanks!

PR #13, opened 2021-12-04
Change name of arguments to share crowdsource results

laurentsimon:
An idea to give an incentive for users to upload their results. Thoughts?

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -35,8 +35,8 @@ inputs:
     description: "INPUT: GitHub token with read access"
     required: True
 
-  share_results:
-    description: "INPUT: share the results with OSSF's public dashboards"
+  badge_support:
+    description: "INPUT: Turn on support for Scorecard badges"
     required: False
 
 branding:
--- a/action.yaml
+++ b/action.yaml
@@ -35,8 +35,8 @@ inputs:
     description: "INPUT: GitHub token with read access"
     required: True
 
-  badge_support:
-    description: "INPUT: Turn on support for Scorecard badges"
+  publish_results:
+    description: "INPUT: Publish results"
     required: False
 
 branding:


laurentsimon:
An idea to give an incentive for users to upload their results. Thoughts?

naveensrinivasan (approved):
Thanks!

action.yaml
```diff
@@ -35,8 +35,8 @@ inputs:
     description: "INPUT: GitHub token with read access"
     required: True
 
-  share_results:
-    description: "INPUT: share the results with OSSF's public dashboards"
+  badge_support:
```
azeemshaikh38, line 38: Ok to submit this PR. 2 questions:

1. Do we have agreement/design that confirms that Actions data will actually be used for Scorecard Badge?
2. Will this data only be used for badge support and nothing else? What if we use this for BQ too?

azeemshaikh38 (approved):

action.yaml
laurentsimon: > Ok to submit this PR. 2 questions:
> 
> 1. Do we have agreement/design that confirms that Actions data will actually be used for Scorecard Badge?

not yet, but it's the plan.

> 2. Will this data only be used for badge support and nothing else? What if we use this for BQ too?

I see it used for everything as ground truth. Note that we have to decide if we want to expose this option for v4 release, or for v5 when implementation is sorted. I think it's worth having it for v4 and jut have a comment that it's not supported yet.

wdut?

azeemshaikh38: I agree that we should expose it in v4. Point being if we plan to expose it for other things later, it'll be sneaky to use a bool named `badge_support` for something else. Else, we update the bool name later and break the folks already using it. So, maybe it should be something more generic (like the existing `share_results` or `publish_results`) which can have a comment saying currently needed for badge support but will be used for other things later.

laurentsimon: good point. I like `publish_result`. @olivekl can you think of a way to explain that publishing will make the data public and will be exposed to our BQ table? You already provided. some suggestion, but I wonder if we can adjust it.

olivekl: Am I understanding correctly that setting `true` for `publish_results` means that users agree that data from their Scorecard runs can be used both for the BQ data table and to award badges, and possibly for other things in the future? 

Is it only for public repos, or would it apply to private repos too?

laurentsimon: > Am I understanding correctly that setting `true` for `publish_results` means that users agree that data from their Scorecard runs can be used both for the BQ data table and to award badges, and possibly for other things in the future?

correct.

> Is it only for public repos, or would it apply to private repos too?

it would apply to private repos too, but users would set this variable to `false` in this case, I think. wdut?

PR #16, opened 2021-12-06
✨ Remove policy from action input

laurentsimon:
This PR removes the policy files as an input to the GitHub action. This will simplify the introduction of scorecard badges which we plan to build on top of raw results. As pointed at by @azeemshaikh38 , having scores for the action and raw results for the badges would make it difficult for users to map one to the other. In addition, most users keep the scores set to 10 for all checks in the policy files: it's easy to ignore. or set as "won't fix" in the GitHub scanning dashboard anyway.


laurentsimon:
cc @chrismcgehee

azeemshaikh38 (changes requested):
Let's hold on this PR till we have an offline discussion.

laurentsimon:
friendly ping to LGTM as discussed offline

azeemshaikh38 (approved):

Issue #19, opened 2021-12-09
Add branch_protection_rule to workflow trigger


laurentsimon:
I've just found out that there's a trigger to run workflows when branch protection settings change https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#branch_protection_rule

As part of the broader effort to use scorecard results for attestation (e.g., SLSA provenance) I think we should enable this trigger for our workflow. In the beginning we'll just ignore the trigger in our implementation. But later, we can create an attestation for this event (including the payload). Besides its use for attestation, it can also be used to complement the data we collect via push events into our BQ table.

cc @azeemshaikh38 @asraa wdut?

azeemshaikh38:
This is a great find, thanks @laurentsimon! +1 to supporting this trigger. I don't understand why you say:

> In the beginning we'll just ignore the trigger in our implementation

We can support this from the start right? Basically we should be able to support: `cron`, `push` and `branch_protection_rule` triggers in our very first release no?

laurentsimon:
I meant ignoring the trigger for attestation. And definitely I agree we can enable it to update the SARIF results. It should be enough to only run the `--checks Branch-Protection` on this trigger.

PR #20, opened 2021-12-09
feature: add support for branch protection trigger

laurentsimon:
closes https://github.com/ossf/scorecard-action/issues/19

See title for description.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -28,6 +28,7 @@ export SCORECARD_POLICY_FILE="$INPUT_POLICY_FILE"
 export SCORECARD_RESULTS_FILE="$INPUT_RESULTS_FILE"
 export SCORECARD_RESULTS_FORMAT="$INPUT_RESULTS_FORMAT"
 export SCORECARD_BIN="/scorecard"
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -53,12 +54,19 @@ then
         $SCORECARD_BIN --local . --format "$SCORECARD_RESULTS_FORMAT" --show-details --policy "$SCORECARD_POLICY_FILE" > "$SCORECARD_RESULTS_FILE"
     fi
 else
-    # For push events, we run on the repo.
+    # For other events, we run on the repo.
```
azeemshaikh38, line 57: For simplicity, why not just run a full run? It'll avoid the additional complexity of https://github.com/ossf/scorecard/pull/1379

azeemshaikh38 (approved):

entrypoint.sh
laurentsimon: well, we may try to enable branch protection on PR, and the change in https://github.com/ossf/scorecard/pull/1379 is useful for that too. I think the code change is not too complex, but if you disagree we can potentially revert.

PR #23, opened 2021-12-10
Fixed the broken codeql action

naveensrinivasan:


.github/workflows/codeql-analysis.yml
```diff
@@ -13,7 +13,6 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ main, * ]
```
laurentsimon, line 16: is this going to run on push to all branches?

.github/workflows/codeql-analysis.yml
```diff
@@ -32,9 +31,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [  ]
-        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
-        # Learn more about CodeQL language support at https://git.io/codeql-language-support
+        language: [ 'javascript' ]
```
laurentsimon, line 34: why not golang?

.github/workflows/codeql-analysis.yml
naveensrinivasan: Yes! The previous syntax was wrong.

.github/workflows/codeql-analysis.yml
naveensrinivasan: There isn't a golang file in this repository.

laurentsimon: oops, sorry... I though we were on scorecard repo :-)
If there's bash script and dockerfile support, we could enable it

laurentsimon: actually I'm curious: what does CodeQl check for if we don't tell it which languages to support? Does it detect it automatically? We have only bash script and docker file in this repo.

naveensrinivasan: It checks for the GitHub actions. We need a js script file to check for vulnerabilities.

laurentsimon: OK.

laurentsimon (approved):
Thanks @naveensrinivasan and sorry for these endless questions :-)

naveensrinivasan:
> Thanks @naveensrinivasan and sorry for these endless questions :-)

👍

Issue #35, opened 2021-12-30
🐛 Support for ghcr.io in Action


laurentsimon:
I tried using the image generated by our release workflow.
When using ghcr.io in the [action](https://github.com/ossf/scorecard-action/blob/main/action.yaml#L44):
```
runs:
  using: "docker"
  image: "ghcr.io://ossf/scorecard-action:latest"
```

I got the following error
```
'ghcr.io://ossf/scorecard-action:latest' should be either '[path]/Dockerfile' or 'docker://image[:tag]'
```

When trying 
```
runs:
  using: "ghcr.io"
  image: "ghcr.io://ossf/scorecard-action:latest"
```

I got 
```
Error: System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter ''using: ghcr.io' is not supported, use 'docker', 'node12' or 'node16' instead.')
```

@naveensrinivasan am I doing something wrong or is docker registry the only registry supported?
cc @azeemshaikh38

azeemshaikh38:
See image below. 2 things:

1. Looks like we can use a non Docker hub image if needed.
2. We can directly reference `Dockerfile` and let GitHub actions build and run that image.

 I personally prefer option (2) if that can be done. Wdut?


![image](https://user-images.githubusercontent.com/4825078/147898449-537f0dd0-c197-4af3-a751-9e232de6527c.png)

laurentsimon:
I like option 2 too, which is what I opted for for this https://github.com/ossf/scorecard-action/pull/38. If you're OK with the PR, plz LGTM and I'll cut a release to start testing.

Issue #33, opened 2021-12-29
RELEASE v1 process


laurentsimon:
steps:
1. cut a scorecard release and wait for a container image to be created and tagged with new release. Note the hash of the container as `CH1`. Note: we do not need a scorecard release, we can use any stable version we want.
2. update the hash pin in our [dockerfile](https://github.com/ossf/scorecard-action/blob/main/.github/Dockerfile#L26) to use the container hash `CH1` from step 1. Once the PR is merged, note the GitHub's commit hash as `GH2`.
~~3. [manually trigger](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) the workflow to generate our container image. Note the hash of the container image generated as `CH3`. It can be found [here](https://github.com/ossf/scorecard-action/pkgs/container/scorecard-action) using the manifest's "digest"`.~~
~~4. update the container image hash we use in [action.yaml:L45](https://github.com/ossf/scorecard-action/blob/main/action.yaml#L45), using the hash `CH3` from step 3. Once the PR is merged, note the GitHub's commit hash as `GH4`.~~
5. test the new hash in a test repo we own. If successful, continue.
6. cut release for the action - the hash of the tagged release should be `GH2`.
7. send a PR to [starter-workflows/code-scanning/scorecards.yml](https://github.com/actions/starter-workflows/tree/main/code-scanning/scorecards.yml) to update the hash to `GH2` from step 4.
8. merge a PR to update our documentation's example workflow to use `GH2`. 
9. verify on the market place that the workflow example contains `GH2`. (the marketplace uses main branch)

justaugustus:
Working on capturing this content in https://github.com/ossf/scorecard-action/pull/194.

justaugustus:
From @laurentsimon in https://github.com/ossf/scorecard-action/pull/194#discussion_r855672197:

> I don't think we release on GH registry. I think it's gcr.io/openssf/scorecard instead. Should be accessible by anyone

justaugustus:
Closed with https://github.com/ossf/scorecard-action/pull/194.

PR #38, opened 2021-12-30
✨ Use local dockerfile instead of pre-built image

laurentsimon:
See https://github.com/ossf/scorecard-action/issues/35, it seems only the docker registry is supported. So instead of finding a way to publish there, let's use a local dockerfile. This PR does just that. It also updates the hash of the scorecard's image to the latest in order to cut a pre-release and start testing.

All in all, this also simplifies the release process for the action. I've updated the release process accordingly in https://github.com/ossf/scorecard-action/issues/33

closes https://github.com/ossf/scorecard-action/issues/35


laurentsimon:
friendly ping. Any questions?

action.yaml
```diff
@@ -42,5 +42,5 @@ branding:
 
 runs:
   using: "docker"
-  image: "docker://laurentsimon/scorecard-action:latest"
+  image: "./Dockerfile"
```
azeemshaikh38, line 45: Nit: maybe remove the `./` as shown in the example provided by GitHub.

azeemshaikh38 (approved):

PR #45, opened 2022-01-06
🐛 Adjust maintained score

laurentsimon:
adjust the score for the Maintained check to pass to 1 instead of 10. Score of 10 means there are a large number of commits/activity in the past 90 days. We don't want to show this in the scanning results so long as there is *some* activity.


azeemshaikh38 (approved):
Thanks!

PR #46, opened 2022-01-06
Olivekl new install instructions

olivekl:
New install instructions, including screenshots.


laurentsimon:
> DRAFT, DO NOT MERGE
> 
> Draft of new installation instructions. Sharing for comments before I add screenshots and links.
> 
> Questions: -Is everything accurate and easy to follow? 

I think we could get rid of it. 

> -Do we need to include the "manual" option at the bottom? Will anyone want to set up the action values themselves, or are they useful for any other reason?

I copied this from other actions' README. But other actions out there are not available under the Security tab, so that's why they need it.

@azeemshaikh38 wdut?

README.md
```diff
@@ -3,40 +3,70 @@
...
-### Inputs
+To install the Scorecards GitHub Action, you need to:
 
-| Name | Required | Description |
-| ----- | -------- | ----------- |
-| `result_file` | yes | The file that contains the results. |
-| `result_format` | yes | The format in which to store the results [json \| sarif]. For GitHub's scanning dashboard, select `sarif`. |
-| `repo_token` | yes | PAT token with read-only access. Follow [these steps](#pat-token-creation) to create it. |
-| `publish_results` | recommended | This will allow you to display a badge on your repository to show off your hard work (release scheduled for Q2'22). See details [here](#publishing-results).|
+1) Authenticate by creating a PAT token and saving the token value as a repository secret (new Scorecards users only)
```
azeemshaikh38, line 10: The `new Scorecard users` part is not true. Even existing Scorecard users will have to upload their PAT as a secret. Maybe we can re-word this line as:

1) Upload a PAT as repository secret for the Scorecard workflow to use.

README.md
```diff
@@ -3,40 +3,70 @@
...
+2) Set up the workflow via the GitHub UI
 
-### PAT token creation
-1. Create a PAT token [here](https://github.com/settings/tokens/new) with the following read permissions:
+If you've already used Scorecards manually in the past, you do not need to create a new PAT token or repository secret and can jump to [Workflow Setup](workflow-setup).
```
azeemshaikh38, line 13: Not true, remove this line.

README.md
```diff
@@ -3,40 +3,70 @@
...
         * `repo > public_repo`
         * `admin:org > read:org`
         * `admin:repo_hook > read:repo_hook`
         * `write:discussion > read:discussion`
-    - Create and copy the token.
+        
+2. Copy the token value. 
 
-2. Create a new repository secret at `https://github.com/<org>/<repo>/settings/secrets/actions/new` with the following settings:
+3. Create a new repository secret at https://github.com/<org>/<repo>/settings/secrets/actions/new (replace <org>/<repo> with the names of your organization and repository) with the following settings:
```
azeemshaikh38, line 27: I think this step is best explained here: https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-a-repository

azeemshaikh38:
> > DRAFT, DO NOT MERGE
> > Draft of new installation instructions. Sharing for comments before I add screenshots and links.
> > Questions: -Is everything accurate and easy to follow?
> 
> I think we could get rid of it.
> 
> > -Do we need to include the "manual" option at the bottom? Will anyone want to set up the action values themselves, or are they useful for any other reason?
> 
> I copied this from other actions' README. But other actions out there are not available under the Security tab, so that's why they need it.
> 
> @azeemshaikh38 wdut?

I think its ok to keep the manual part of the setup for more advanced users.

README.md
olivekl: How about this?  

To install the Scorecards GitHub Action, you need to:

1) Create a Personal Access Token (PAT) and save the token value as a repository secret;

(Note: current Scorecards users can reuse their existing PAT for the repository secret)

2) Set up the workflow via the GitHub UI

olivekl:
Thanks, Azeem. I addressed your comments. I'll add screenshots for the UI section if everything else looks good.

README.md
azeemshaikh38: LGTM, thanks!

azeemshaikh38 (approved):

README.md
```diff
@@ -3,40 +3,85 @@
...
+    
+    (Note: current Scorecards users can reuse their existing PAT for the repository secret.)
+    
+3) Set up the workflow via the GitHub UI
 
-### PAT token creation
-1. Create a PAT token [here](https://github.com/settings/tokens/new) with the following read permissions:
+### Authentication
+1. [Create a Personal Access Token](https://github.com/settings/tokens/new) with the following read permissions:
     - Note: `Read-only token for OSSF Scorecard Action`
```
laurentsimon, line 18: since it seems hard to re-use an existing token (UX does not allow reading its value), users may often create a new token. If all tokens have the same name it will be confusing. Shall we tweak the name to include the repo name, e.g., `Read-only token for OSSF Scorecard Action - myorg/myrepo` or `Read-only token for OSSF Scorecard Action - repo:myrepo` ?

README.md
```diff
@@ -3,40 +3,85 @@
 > Official GitHub Action for [OSSF scorecard](https://github.com/ossf/scorecard).
 
 ## Installation
-The simplest and quickest way to install Scorecards's GitHub action is from the [GitHub's marketplace](https://github.com/marketplace/actions/ossf-scorecard-action).
-**WARNING**: the action does *not* support private repositories.
+**NOTE**: The Scorecards GitHub Action does not support private repositories. Private repositories can run Scorecards manually by following the [standard installation instructions)[https://github.com/ossf/scorecard#installation].
```
laurentsimon, line 6: so the news is that we support private repos:
1. running the action works if they follow the permission declaration as in https://github.com/test-organization-ls/action-test/blob/main/.github/workflows/scorecards-analysis.yml#L19
2. upload of the results to the scanning dashboard only work for repos that have GitHub Advanced Security (not under our control)

The use case of 1) would be for a user who wants to run scorecard and do some analysis on their own. Maybe we can ignore this part and just say that the action is supported for private repos with GitHub Advanced Security.

README.md
```diff
@@ -3,40 +3,85 @@
...
 
-| Name | Required | Description |
-| ----- | -------- | ----------- |
-| `result_file` | yes | The file that contains the results. |
-| `result_format` | yes | The format in which to store the results [json \| sarif]. For GitHub's scanning dashboard, select `sarif`. |
-| `repo_token` | yes | PAT token with read-only access. Follow [these steps](#pat-token-creation) to create it. |
-| `publish_results` | recommended | This will allow you to display a badge on your repository to show off your hard work (release scheduled for Q2'22). See details [here](#publishing-results).|
+1) Create a Personal Access Token (PAT) for authentication and save the token value as a repository secret; 
+    
+    (Note: current Scorecards users can reuse their existing PAT for the repository secret.)
```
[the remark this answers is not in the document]
olivekl: Yes, I think it's better to just remove all references to reusing the PAT, given what you point out. I was able to reuse mine because I had saved it back when I first created it, but I realize this could introduce another point of confusion.

README.md
```diff
@@ -3,40 +3,70 @@
 > Official GitHub Action for [OSSF scorecard](https://github.com/ossf/scorecard).
 
 ## Installation
-The simplest and quickest way to install Scorecards's GitHub action is from the [GitHub's marketplace](https://github.com/marketplace/actions/ossf-scorecard-action).
-**WARNING**: the action does *not* support private repositories.
+**NOTE**: The Scorecards GitHub Action does not support private repositories. Private repositories can run Scorecards manually by following the [standard installation instructions)[https://github.com/ossf/scorecard#installation].
```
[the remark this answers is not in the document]
olivekl: I actually meant manually to mean the command line, but given your question about it I'll be more specific. 
I have been reusing the same token I had used on the command line so I can confirm that it works. Maybe we should ask people to create a new token to cut down on confusion about not being able to find it if they haven't saved it, though. Is there any drawback to creating more tokens?

README.md
olivekl: Updated: " (Note: If you have already installed Scorecards on your repository from the command line, you can reuse your existing PAT for the repository secret. If you not longer have access to the PAT, though, simply create a new one.)"

Hopefully clears up all confusion?

README.md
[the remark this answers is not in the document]
olivekl: I've changed it to specify "from the command line" instead of manually. 

But for the token name: is it the name of the token that matters, or the name of the repository secret where you copy-paste the value of the token? I think you mean the latter, right?

README.md
olivekl: That makes sense. Do we expect people to want to install Scorecards on many repos, say dozens or hundreds? This step could get onerous if so, but I'm not sure whether that's a reasonable expected use case for the target user.

README.md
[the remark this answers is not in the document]
laurentsimon: I don't think so. Let's suggest the easiest solution.

README.md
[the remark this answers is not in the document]
laurentsimon: you're correct, name of repository secret

README.md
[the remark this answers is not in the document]
laurentsimon: agreed. Let's remove PAR re-use

README.md
laurentsimon: I expect some users to use it on multiple repos, but there's no tool to scale it to hundreds of repos atm. There may be a way to automate it via a tool we write (create token, save as secrets and create pull requests), but it does not exist today

README.md
olivekl: Yeah, I agree, I doubt there's a lot of users falling into that use case. Let's keep it simple for now.

README.md
laurentsimon: fyi, we do support private repositories now. `publish_results` is set to false and the value provide by developers in the workflow is ignored. Let's update the example workflow and the doc to reflect it?

README.md
```diff
@@ -2,41 +2,84 @@
...
     - Scopes: 
         * `repo > public_repo`
         * `admin:org > read:org`
         * `admin:repo_hook > read:repo_hook`
         * `write:discussion > read:discussion`
-    - Create and copy the token.
 
-2. Create a new repository secret at `https://github.com/<org>/<repo>/settings/secrets/actions/new` with the following settings:
-    - Name: `SCORECARD_TOKEN`
+![image](/images/tokenscopes.png)
```
laurentsimon, line 26: please update the name of the token in the example workflow to `SCORECARD_READ_TOKEN `

laurentsimon: I think this will be automatically updated when you rebase from main.

README.md
laurentsimon: fyi, this is the text I'm using for the default workflow https://github.com/actions/starter-workflows/pull/1345/files

olivekl: Done.

README.md
```diff
@@ -3,40 +3,85 @@
...
+
+
+4) Commit the changes.
+![image](/images/install04.png)
+
+
+The workflow is preconfigured to run on every repository contribution. Results are available on the GitHub code-scanning dashboard[TODO:ADD LINK?] for remediation. 
+
+## Verify Runs and View Results
+To verify that the Action is running successfully, click the repository's Actions tab to see the status of all recent workflow runs.
```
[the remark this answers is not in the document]
olivekl: Good idea, done.

README.md
```diff
@@ -3,40 +3,70 @@
...
+
+3) Commit and merge the changes.
+
+The workflow is preconfigured to run on every repository contribution. Results are available on the GitHub code-scanning dashboard[TODO:ADD LINK?] for remediation. 
+
+## Verify Runs and View Results
+To verify that the Action is running successfully, click the repository's Actions tab to see the status of all recent workflow runs. 
+       
+To view a list of results from each Scorecards Action run, go to the Security tab and click "Code Scanning Alerts." 
+
```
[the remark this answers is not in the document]
olivekl: Done

README.md
```diff
@@ -3,40 +3,85 @@
...
+Note: if you have aleady configured other code scanning tools, your UI will look different than shown above. Instead, click "Code Scanning Alerts" on the left side of the page. 
+
+![image](/images/installb1.png)
+
+Then click "Add More Scanning Tools."
+
+![image](/images/installb2.png)
+
+3) Choose the "OSSF Scorecards supply-chain security analysis" from the list of workflows, and then click “set up this workflow.”
```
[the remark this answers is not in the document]
laurentsimon: see below:

Screen Shot 2022-01-11 at 12 15 29 PM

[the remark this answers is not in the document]
olivekl: How is it now?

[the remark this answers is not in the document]
laurentsimon: It works. Actually it looks better than most other images that are very close to the sentence.

laurentsimon (approved):
Thank you!

laurentsimon:
We just need to rebase the PR.

PR #58, opened 2022-01-14
Fixing repo token secret typo

abirismyname:
New Starter Workflow refers to SCORECARD_READ_TOKEN but README uses SCORECARD_TOKEN. Making the variable name consistent.


laurentsimon (approved):
oops, this slipped thru the cracks. We updated the text but not the example workflow. Thank you for catching this!

PR #60, opened 2022-01-14
Bump dockerfile hash to fix SARIF semVer

laurentsimon:
See https://github.com/ossf/scorecard/issues/1479 for details

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@
 #           -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN \
 #           -e GITHUB_REPOSITORY="ossf/scorecard" \
 #           laurentsimon/scorecard-action:latest
-FROM gcr.io/openssf/scorecard@sha256:a6db69ae36f3199ada709fe86dd9a226d75fa1d20397326006473f54d60f4165 as base
+FROM gcr.io/openssf/scorecard@sha256:8165ad910019422f40c51cbb97ff6e7db0e2e2e11faebf59e0b6f1a2eb66ebd7 as base
 
 # Build our image and update the root certs.
 # TODO: use distroless.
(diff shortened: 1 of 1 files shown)


laurentsimon:
FYI, need to update the hash with the new v4.0.1 scorecard release.

Dockerfile
```diff
@@ -21,7 +21,7 @@
 #           -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN \
 #           -e GITHUB_REPOSITORY="ossf/scorecard" \
 #           laurentsimon/scorecard-action:latest
-FROM gcr.io/openssf/scorecard@sha256:7def80b7386d0addfa20b408fe23eacf85280806dc1981782f4e7766c2eb8833 as base
+FROM gcr.io/openssf/scorecard@sha256:a6db69ae36f3199ada709fe86dd9a226d75fa1d20397326006473f54d60f4165 as base
```
azeemshaikh38, line 24: Just cut a `v4.0.1` release on Scorecard. Will wait to verify that the new hash is unchanged before approving.

laurentsimon: I will change since we don't have reproducible builds yet. Also, the version embedded in scorecard binary will be bumped.

laurentsimon: new hash is `8165ad910019422f40c51cbb97ff6e7db0e2e2e11faebf59e0b6f1a2eb66ebd7`

azeemshaikh38: New image here: `gcr.io/openssf/scorecard@sha256:8165ad910019422f40c51cbb97ff6e7db0e2e2e11faebf59e0b6f1a2eb66ebd7`

laurentsimon: updated in the PR. PTAL

azeemshaikh38 (approved):

Issue #73, opened 2022-01-28
`Only the default branch 'refs/heads/null' is supported` after #68 landed


ChrisCarini:
Hello, I am getting the below error after #68 (fix for #66) landed and was picked up by my workflow:

```
Event file: /github/workflow/event.json
Event name: schedule
Ref: refs/heads/master
Private repository: null
Publication enabled: true
Format: sarif
Policy file: /policy.yml
Default branch: refs/heads/null
refs/heads/master not supported with 'schedule' event.
Only the default branch 'refs/heads/null' is supported
```
(**ref:** [link](https://github.com/ChrisCarini/environment-variable-settings-summary-intellij-plugin/runs/4955265412?check_suite_focus=true))

This is happening on the scheduled runs of the workflow I have configured.

@laurentsimon - when you have a moment to take a look, could you help investigate? The repo in which this is run has the default branch set to `master`.

Thank you!

laurentsimon:
Thanks. I think I know what's happening. The `schedule` trigger does not seem to have an associated branch, which is unexpected from me. I'll work on a fix.

cc @naveensrinivasan let's add better unit test for the action.

laurentsimon:
fyi, I've created a support ticket with GitHub.

ChrisCarini:
Awesome, thank you @laurentsimon !

ChrisCarini:
SGTM, and no worries! Appreciate you looking into this (and #84 😄 )!

PR #75, opened 2022-01-29
🐛 Fix null default branch/private fields

laurentsimon:
The `repository.private` and `repository.default_branch` are null when the action is triggered by a `schedule` trigger.
I think this is a bug and I've created a support issue on GitHub to ask.

There is another field `master_branch` but it may be deprecated in the future so I would like to avoid it.

To fix the issue, I'm querying a REST API instead, which seems more reliable.

closes https://github.com/ossf/scorecard-action/issues/73

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -30,15 +30,27 @@ export SCORECARD_POLICY_FILE="/policy.yml" # Copied at docker image creation.
 export SCORECARD_RESULTS_FILE="$INPUT_RESULTS_FILE"
 export SCORECARD_RESULTS_FORMAT="$INPUT_RESULTS_FORMAT"
 export SCORECARD_PUBLISH_RESULTS="$INPUT_PUBLISH_RESULTS"
-# https://docs.github.com/en/actions/learn-github-actions/environment-variables
 export SCORECARD_PRIVATE_REPOSITORY="$(jq '.repository.private' $GITHUB_EVENT_PATH)"
 export SCORECARD_DEFAULT_BRANCH="refs/heads/$(jq -r '.repository.default_branch' $GITHUB_EVENT_PATH)"
 export SCORECARD_BIN="/scorecard"
 export ENABLED_CHECKS=
 
-export SCORECARD_MASTER_BRANCH="refs/heads/$(jq -r '.repository.master_branch' $GITHUB_EVENT_PATH)"
-
-# WARNING: boolean inputs are strings https://github.com/actions/runner/issues/1483.
+## ============================== WARNING ======================================
+# https://docs.github.com/en/actions/learn-github-actions/environment-variables
+# export SCORECARD_PRIVATE_REPOSITORY="$(jq '.repository.private' $GITHUB_EVENT_PATH)"
+# export SCORECARD_DEFAULT_BRANCH="refs/heads/$(jq -r '.repository.default_branch' $GITHUB_EVENT_PATH)"
+#
+# The $GITHUB_EVENT_PATH file produces:
+# private: null
+# default_branch: null
+#
+# for trigger event `schedule`. This is a bug.
+# So instead we use the REST API instead to retrieve the data.
+#
+# Boolean inputs are strings https://github.com/actions/runner/issues/1483.
+# ===============================================================================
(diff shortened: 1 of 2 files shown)


laurentsimon:
@ChrisCarini please verify that this fixes the problem. Use `uses: laurentsimon/scorecard-action@bug/defaultbranch` temporarily.

I'll push the changes/new version one you've confirmed. 

Thanks for your patience!

azeemshaikh38:
> @ChrisCarini please verify that this fixes the problem. Use `uses: laurentsimon/scorecard-action@bug/defaultbranch` temporarily.
> 
> I'll push the changes/new version one you've confirmed.
> 
> Thanks for your patience!

We need to find a better way to do integration tests. Could we prioritize it for this quarter?

azeemshaikh38 (approved):

laurentsimon:
> > @ChrisCarini please verify that this fixes the problem. Use `uses: laurentsimon/scorecard-action@bug/defaultbranch` temporarily.
> > I'll push the changes/new version one you've confirmed.
> > Thanks for your patience!
> 
> We need to find a better way to do integration tests. Could we prioritize it for this quarter?

yes, I totally agree. @naveensrinivasan and I will chat next week. There's some really simply low-hanging fruit we can do in the next couple weeks. I'll write an issue with a list of test we can start with.

entrypoint.sh
```diff
@@ -30,13 +30,26 @@ export SCORECARD_POLICY_FILE="/policy.yml" # Copied at docker image creation.
...
+# private: null
+# default_branch: null
+#
+# for trigger event `schedule`. This is a bug.
+# So instead we use the REST API to retrieve the data.
+#
+# Boolean inputs are strings https://github.com/actions/runner/issues/1483.
+# ===============================================================================
+export SCORECARD_PRIVATE_REPOSITORY="$(curl -s -H \"Authorization: Bearer $GITHUB_AUTH_TOKEN\" https://api.github.com/repos/$SCORECARD_REPOSITORY | jq -r '.private')"
+export SCORECARD_DEFAULT_BRANCH="refs/heads/$(curl -s -H \"Authorization: Bearer $GITHUB_AUTH_TOKEN\" https://api.github.com/repos/$SCORECARD_REPOSITORY | jq -r '.default_branch')"
```
ChrisCarini, line 52: Might I recommend something like below instead:
```suggestion
curl -s -H "Authorization: Bearer $GITHUB_AUTH_TOKEN" https://api.github.com/repos/$SCORECARD_REPOSITORY > repo_info.json
export PR="$(cat repo_info.json | jq -r '.private')"
export DB="refs/heads/$(cat repo_info.json | jq -r '.default_branch')"
rm repo_info.json
```

This reduces the API requests by 1 and just extracts the needed information from a file. I've run into API rate-limiting issues calling the GH API from within GitHub Actions (I am an GH action author, myself), so I figured this might be a nice improvement/optimization for this action, as well.

ChrisCarini (approved):
**LGTM!** Just a suggestion to reduce the calls - please consider accepting the suggestion into this PR! 😄 

### Brief testing notes below:

1.  Switched workflow to use @laurentsimon 's bug branch ([commit ref](https://github.com/ChrisCarini/environment-variable-settings-summary-intellij-plugin/commit/4250e3553ddb2d872c5a3f43a6444d70c66036c9))
2. Observe results
    1. **`push` workflow execution:** ✅ Success ([ref](https://github.com/ChrisCarini/environment-variable-settings-summary-intellij-plugin/runs/4994493523?check_suite_focus=true))
    2. **`workflow_dispatch` workflow execution for `master` branch (default branch of this repo):** ✅ Success ([ref](https://github.com/ChrisCarini/environment-variable-settings-summary-intellij-plugin/runs/4994499321?check_suite_focus=true))
    3. **`workflow_dispatch` workflow execution for `addIconDir` branch (non-default branch):** ✅ Failure (expected; non-default branch not supported per failure message) ([ref](https://github.com/ChrisCarini/environment-variable-settings-summary-intellij-plugin/runs/4994549068?check_suite_focus=true))

entrypoint.sh
laurentsimon: Good call! Will do.

PR #76, opened 2022-01-29
Update codeql-analysis.yml

jauderho:
Given the nature of the scorecard action, the actions within the scorecard-action should be using commit hashes.


laurentsimon (approved):
Thank you! I created the workflow thru GitHub and it automatically generated this one :/
Good catch!

PR #82, opened 2022-01-31
Update Dockerfile and add dependabot support for Docker

jauderho:
Noticed that the Dockerfile is using an old version of stretch. Current version is 9.13 not 9.5.

This PR adds dependabot support for updating Dockerfile as well as converting the existing Dockerfile to use a digest for the debian:9.5-slim image.

Once this PR is accepted and merged, dependabot should create a new PR to update to debian:9.13-slim.

However, given that support for Stretch ends in about 5 months according to https://endoflife.date/debian, consideration should be given towards a new Debian version but I will leave that to the maintainers to decide.


Dockerfile
```diff
@@ -25,7 +25,8 @@ FROM gcr.io/openssf/scorecard@sha256:8165ad910019422f40c51cbb97ff6e7db0e2e2e11fa
 
 # Build our image and update the root certs.
 # TODO: use distroless.
-FROM debian:9.5-slim
+#FROM debian:9.5-slim
```
laurentsimon, line 28: can you add `# debian:9.5-slim` as comment right side of the hash?

loosebazooka: I think if you add the tag into the image reference, dependabot can pick it up

debian:9.5-slim@sha256:ef6be890318a105f7401d0504c01c2888daa5d9e45b308cf0e45c7cb8e44634f

loosebazooka: also 9.5-slim is crazy old - I would use 11.2-slim or something (11 is current stable debian, 9 has been deprecated for a while)

jauderho: I wasn't aware of this possible format before and wanted to verify it in one of my repos. Looks like dependabot does the right thing and fixes up both the version and digest at the same time: https://github.com/jauderho/sandbox/pull/127/files

Nice!

jauderho:
I have fixed the dependabot.yml so that it'll merge cleanly.

Issue #79, opened 2022-01-31
Feature: Improve test coverage for scorecard-action


laurentsimon:
Things we should be able to test:

**=== Unit tests: P0 =====**
1. private repos
2. different trigger events: push, schedule, pull_request, workflow_dispatch, etc. Priority should be for push, schedule and pull request
3. different branches: we only support the default branch for push and schedule 
3. forks (not supported), should be added to documentation https://github.com/ossf/scorecard-action/issues/80
4. this problem https://github.com/ossf/scorecard-action/issues/106#issuecomment-1049247360
5. Capture the different threshold scores, e.g. https://github.com/ossf/scorecard-action/pull/129

The entrypoint of the action is entrypoint.sh. It expects some variables to be set, see https://github.com/ossf/scorecard-action/blob/main/entrypoint.sh#L25. These variables are the one we need to change for each unit test.

The shell script is part of a dockerfile: the dockefile defines the GitHub action, so we can build an image and call it with different parameters to create our unit tests.

The file `$GITHUB_EVENT_PATH` should contain all the data we want, except for a GitHub API bug, see https://github.com/ossf/scorecard-action/blob/main/entrypoint.sh#L38. his means we may need to create severals repos for testing purposes until the issue is resolved. 
https://github.com/ossf/scorecard-action/pull/75#pullrequestreview-867031359

Here is an example of tests done by a contributor who filed an issue https://github.com/ossf/scorecard-action/pull/75#pullrequestreview-867031359

**=== Unit tests: P1 =====**
1. https://github.com/ossf/scorecard/issues/1196 should test that command line does not change; and that the same checks are enabled with the same user's input

**=== End-to-end tests: P2 ====**
Run the workflow/upload the SARIF results to a repo and retrieve the results via an API. Verify the results shown are as expected. This is low priority, I think, because we already test SARIF generation in scorecard itself.

cc @azeemshaikh38 @naveensrinivasan anything missing?

naveensrinivasan:
Discussed offline with @laurentsimon, Instead of writing tests for the shell script. I am proposing we move the shell script into a `golang` code, which would be easier to maintain and specifically write `tests`.

azeemshaikh38:
[...]

PR #87, opened 2022-02-04
:sparkles:  Initial porting the shellscript to go

naveensrinivasan:
Porting the shellscript to go

related #79


main.go
```diff
@@ -0,0 +1,157 @@
...
+func initalizeENVVariables() error {
+	/*
+	 https://docs.github.com/en/actions/learn-github-actions/environment-variables
+	   GITHUB_EVENT_PATH contains the json file for the event.
+	   GITHUB_SHA contains the commit hash.
+	   GITHUB_WORKSPACE contains the repo folder.
+	   GITHUB_EVENT_NAME contains the event name.
+	   GITHUB_ACTIONS is true in GitHub env.
+	*/
+	if err := os.Setenv("ENABLE_SARIF", "1"); err != nil {
```
laurentsimon, line 43: wan to create a set of errors we can re-use?

laurentsimon (approved):
Thanks, looking good

main.go
naveensrinivasan: Can I do it as part of the next PR?

laurentsimon: sure!

azeemshaikh38: +1. Will be useful to have re-usable fns here. Maybe consider a `map[string]string` which can be iterated over to set the envVars. Can do this in next PR.

azeemshaikh38 (approved):
Thanks @naveensrinivasan!

main.go
naveensrinivasan: Sounds good. Thanks, guys!

Issue #52, opened 2022-01-11
Creat tool to enable action on multiple repos


laurentsimon:
Organizations and users may want to enable an action at scale. This requires 
1. creating a PAT 
2. saving as secrets in each repo
3. Filing pull requests

cc @inferno-chromium 

@josepalafox is there an org setting to enable a workflow at scale?

laurentsimon:
cc @azeemshaikh38

justaugustus:
I'm going to take over cleanup of the multi-repo-action.

azeemshaikh38:
@justaugustus are you still planning to work on this? Would be good to announce this tool along with our v2 launch (aiming for a launch timeline of first week of September). Let us know if you need help or are unable to work on this.

raghavkaul:
You could probably automate several of the setup steps using something like [Terraform for GitHub](https://registry.terraform.io/providers/integrations/github/).

justaugustus:
> @justaugustus are you still planning to work on this? Would be good to announce this tool along with our v2 launch (aiming for a launch timeline of first week of September). Let us know if you need help or are unable to work on this.

Yep, I can work on this before then!

azeemshaikh38:
Awesome thanks! Looking forward to it :)

justaugustus:
Continuing the cleanup work in https://github.com/ossf/scorecard-action/pull/833.

justaugustus:
@lelia has been working on some cleanups and testing them on Cisco repositories, so I'll assign this to her.

PR #88, opened 2022-02-05
enable workflow at organization level

rohankh532:
#52


azeemshaikh38:
Thanks for the quick turnaround here @rohankh532! Looks good overall, will take another look next week. Please include documentation for the tool in this PR (things like how to use the tool, required scopes for the PAT etc.). Could add to `README.md` so that it's visible to users.

multi-repo-action/go.mod
```diff
@@ -0,0 +1,18 @@
+module multi_repo_action.go
```
laurentsimon, line 1: how about calling this `cli` for simplicity?

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+	)
+	tokenClient := oauth2.NewClient(context, tokenService)
+	client := github.NewClient(tokenClient)
+
+	//Get repositories under organization
+	lops := &github.RepositoryListByOrgOptions{Type: "all"}
+	repos, _, err := client.Repositories.ListByOrg(context, ORG_NAME, lops)
+	err_check(err, "List Org Repos Error")
+
+	//Convert to list of repository names
```
laurentsimon, line 33: Try writing comments as: `// Comment starts with space and ends with full dot.`

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+	repoNames := []string{}
+	for _, repo := range repos {
+		repoNames = append(repoNames, *repo.Name)
+	}
+
+	//Get yml file into byte array
+	fileContent, _ := ioutil.ReadFile("scorecards.yml")
```
laurentsimon, line 40: check the error.

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+		//Get head commit SHA of default branch
+		repo, _, err := client.Repositories.Get(context, ORG_NAME, repoName)
+		err_check(err, "Get Repository Error")
+		defaultBranch, _, err := client.Repositories.GetBranch(context, ORG_NAME, repoName, *repo.DefaultBranch, true)
+		err_check(err, "Get Branch Error")
+		defaultBranchSHA := defaultBranch.Commit.SHA
+
+		//Create new branch
+		ref := &github.Reference{
+			Ref:    github.String("refs/heads/scorecard"),
```
laurentsimon, line 54: what if this branch already exists?

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/google/go-github/v42/github"
+	"golang.org/x/oauth2"
+)
+
+// *****SET THESE PARAMETERS*****
+const ORG_NAME string = "organization name"
```
laurentsimon, line 15: orgs/individual may want to install only on a subset of the repos. This is something that was requested, so let's add support for it.

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+		_, _, err = client.Git.CreateRef(context, ORG_NAME, repoName, ref)
+		err_check(err, "Create Ref Error")
+
+		//Create file in repository
+		opts := &github.RepositoryContentFileOptions{
+			Message: github.String("Adding scorecard workflow"),
+			Content: fileContent,
+			Branch:  github.String("scorecard"),
+		}
+		_, _, err = client.Repositories.CreateFile(context, ORG_NAME, repoName, ".github/workflows/scorecards.yml", opts)
```
laurentsimon, line 66: name is scorecards-analysis.yml, to be consistent with he doc https://github.com/ossf/scorecard-action#manual-action-setup

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+		err_check(err, "CreateFile Error")
+
+		//Wait for workflow file to finish creating
+		time.Sleep(3 * time.Second)
```
laurentsimon, line 70: is this needed because the `CreateFile()` function crates the remote branch?
I wonder if there's a more reliable way to do this. Network latency may vary, etc. Let's try to have a more robust way to catch this error and convey it to user?

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+			Object: &github.GitObject{SHA: defaultBranchSHA},
+		}
+		_, _, err = client.Git.CreateRef(context, ORG_NAME, repoName, ref)
+		err_check(err, "Create Ref Error")
+
+		//Create file in repository
+		opts := &github.RepositoryContentFileOptions{
+			Message: github.String("Adding scorecard workflow"),
+			Content: fileContent,
+			Branch:  github.String("scorecard"),
```
laurentsimon, line 64: what if this branch already exists?
Can you add a comment about the difference between `Branch` and `Ref` field above?

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+		err_check(err, "CreateFile Error")
+
+		//Wait for workflow file to finish creating
+		time.Sleep(3 * time.Second)
+
+		//Create Pull request
+		pr := &github.NewPullRequest{
+			Title: github.String("Added Scorecard Workflow"),
+			Head:  github.String("scorecard"),
+			Base:  github.String(*defaultBranch.Name),
```
laurentsimon, line 76: verify the pointer is not `nil`

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+			Body:  github.String("Added the workflow for Google's OSSF Security Scorecard"),
```
laurentsimon, line 77: remove Google. `OpenSSF's Scorecards` should be enough.

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+	//Convert to list of repository names
+	repoNames := []string{}
+	for _, repo := range repos {
+		repoNames = append(repoNames, *repo.Name)
+	}
+
+	//Get yml file into byte array
+	fileContent, _ := ioutil.ReadFile("scorecards.yml")
+
+	for _, repoName := range repoNames {
```
laurentsimon, line 42: Something to think about: if the PR fails for some reasons, the user will be left in a state where *some* repos only will have the PR created. How do we recover? How do we allow users to run the command but not create duplicated PRs for repos that already have it created?

multi-repo-action/scorecards.yml
```diff
@@ -0,0 +1,58 @@
+name: Scorecards supply-chain security
```
laurentsimon, line 1: We need to keep this workflow up-to-date, in particular the hashes. Ideally we should use the hash for the latest version of scorecard and set it dynamically to `ossf/scorecard-action@hash`. Try thinking of a way to do this. Otherwise it's yet another file to keep up-to-date manually.

Maybe to keep things simple: we could update the release process in https://github.com/ossf/scorecard-action/issues/33. The only downside of this is that if a user clones the repo, they won't get the update unless they `git pull`.

Note that an out-of-date hash is not a super big deal in practice: dependabot will notice the hash is old and create a PR to update it. So I think updating the https://github.com/ossf/scorecard-action/issues/33 is the way to do.

So I just ignore everything I said :-)

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+		_, _, err = client.PullRequests.Create(context, ORG_NAME, repoName, pr)
+		err_check(err, "Pull Request Error")
+
+		//Logging
+		fmt.Println("Added scorecard workflow PR from scorecard to", *defaultBranch.Name, "branch of repo", repoName)
+	}
+}
+
+func err_check(err error, msg string) {
+	if err != nil {
```
laurentsimon, line 90: +1, this is great!
@naveensrinivasan maybe we could do this for the golang version of the action to make the code more readable?

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,94 @@
...
+	"os"
+	"time"
+
+	"github.com/google/go-github/v42/github"
+	"golang.org/x/oauth2"
+)
+
+// *****SET THESE PARAMETERS*****
+const ORG_NAME string = "organization name"
+const PAT string = "access token"
```
laurentsimon, line 16: his PAT will only need read permission, is this correct?

I wonder if there's a way to do this in a workflow, to avoid having users install golang, etc. My question I think is: can the default `secrets.GITHUB_TOKEN` be used? This way, the org maintainers need not create a PAT, need to have to revoke it, etc.

Just something to keep in mind, not this PR :-)

laurentsimon (commented):
Thanks @rohankh532 this looks pretty good.
Just fyi, we tend to comment a lot in general, this is entirely normal ^^

multi-repo-action/org-workflow-add.go
rohankh532: defaultBranch is verified at the start of the loop

multi-repo-action/org-workflow-add.go
rohankh532: Made errors skip to next repo in iteration instead of doing os.Exit

multi-repo-action/org-workflow-add.go
rohankh532: added check above

multi-repo-action/org-workflow-add.go
rohankh532: Yeah I tested and it's not needed. This is a remnant from earlier when I was triggering the workflow after creating the file.

multi-repo-action/org-workflow-add.go
rohankh532: How do we define the subset? Is it by a certain property, or should we allow them to provide a list of repos?

multi-repo-action/org-workflow-add.go
rohankh532: Yes, only required permissions are public_repo and I think also read:org?

naveensrinivasan (dismissed):
Without tests, it is hard to maintain things. For example that is the reason we are porting the shell script to `go` #87

We shouldn't merge these changes without tests and high code coverage. 


That is one of the reasons I am blocking this change.

multi-repo-action/org-workflow-add.go
laurentsimon: there's another case: repos where scorecard is not installed but you already sent a PR. How do you skip these repos?

multi-repo-action/org-workflow-add.go
laurentsimon: I suppose a list of repo, or a config file. feel free to suggest.

multi-repo-action/org-workflow-add.go
laurentsimon: interesting tools I came across:
https://github.com/lindell/multi-gitter
https://cli.github.com/manual/gh_extension

Not saying we should these, just sending links for inspiration and fyi.

azeemshaikh38:
> Without tests, it is hard to maintain things. For example that is the reason we are porting the shell script to `go` #87
> 
> We shouldn't merge these changes without tests and high code coverage.
> 
> That is one of the reasons I am blocking this change.

@naveensrinivasan do you have any suggestions on how to unit test this? Since the code is mostly API calls to GH, it's difficult to write unit tests for it.

multi-repo-action/org-workflow-add.go
rohankh532: Added repo list as a parameter

naveensrinivasan:
> > Without tests, it is hard to maintain things. For example that is the reason we are porting the shell script to `go` #87
> > We shouldn't merge these changes without tests and high code coverage.
> > That is one of the reasons I am blocking this change.
> 
> @naveensrinivasan do you have any suggestions on how to unit test this? Since the code is mostly API calls to GH, it's difficult to write unit tests for it.

I agree unit tests could be hard for this. But we should be able to add `e2e` tests for this similar to https://github.com/ossf/scorecard/tree/main/e2e . This ensures that we haven't broken the functionality especially when there are lots of folks going to use this as a feature

azeemshaikh38:
> > > Without tests, it is hard to maintain things. For example that is the reason we are porting the shell script to `go` #87
> > > We shouldn't merge these changes without tests and high code coverage.
> > > That is one of the reasons I am blocking this change.
> > 
> > 
> > @naveensrinivasan do you have any suggestions on how to unit test this? Since the code is mostly API calls to GH, it's difficult to write unit tests for it.
> 
> I agree unit tests could be hard for this. But we should be able to add `e2e` tests for this similar to https://github.com/ossf/scorecard/tree/main/e2e . This ensures that we haven't broken the functionality especially when there are lots of folks going to use this as a feature

Makes sense. Since that is a big chunk of work in itself, can we tackle that as part of a separate PR? Note that, this PR is just a v0 version of the tool, something to get the framework in place. It will get refined more as we go before we actually recommend this to users.

multi-repo-action/go.mod
```diff
@@ -0,0 +1,17 @@
+module cli
```
naveensrinivasan, line 1: Could the module name be something like this  `github.com/ossf/scorecard-actions/cli`?

https://github.com/ossf/scorecard/blob/6930c3ab3b715808a0d28d383c73b151bb495fe2/tools/go.mod#L1

multi-repo-action/org-workflow-add.go
rohankh532: Added checks so the repo gets skipped if the branch or workflow already exists. I didn't check for the PR existing because if the base branch doesn't exist, then the PR can't.

azeemshaikh38:
> It could be a separate PR. But the issue is that no one takes the responsibility and it falls off the radar till things go wrong. That is the reason we should insist on tests for PRs.
> 
> Also writing tests make the code more manageable into functions instead of a giant main func.
> 
> Thoughts?

Thinking about this more - it won't be that easy writing e2e tests also for this. Every e2e test will end up creating a bunch of new PRs and unlike Scorecards, we won't have data that is being returned to verify against. I do genuinely think writing tests for this is far more complicated than what we want to take on in this initial PR. If you still have concerns, let's discuss this in the upcoming Scorecard sync.

multi-repo-action/scorecards.yml
rohankh532: Made it so the latest workflow file from https://github.com/ossf/scorecard/blob/main/.github/workflows/scorecard-analysis.yml is pulled

multi-repo-action/org-workflow-add.go
rohankh532: Added check

multi-repo-action/scorecards.yml
laurentsimon: maybe we could use a workflow template committed to this repo, and use ascii doc to include the template into the README. The scorecard workflow has lots of comments tat we should avoid in the PR.

rohankh532: So the scorecard-analysis.yml file that gets created should have no comments, and the original version gets put in the README? The yml doesn't seem to have that many comments though.

rohankh532:
> > > > > Without tests, it is hard to maintain things. For example that is the reason we are porting the shell script to `go` #87
> > > > > We shouldn't merge these changes without tests and high code coverage.
> > > > > That is one of the reasons I am blocking this change.
> > > > 
> > > > 
> > > > @naveensrinivasan do you have any suggestions on how to unit test this? Since the code is mostly API calls to GH, it's difficult to write unit tests for it.
> > > 
> > > 
> > > I agree unit tests could be hard for this. But we should be able to add `e2e` tests for this similar to https://github.com/ossf/scorecard/tree/main/e2e . This ensures that we haven't broken the functionality especially when there are lots of folks going to use this as a feature
> > 
> > 
> > Makes sense. Since that is a big chunk of work in itself, can we tackle that as part of a separate PR? Note that, this PR is just a v0 version of the tool, something to get the framework in place. It will get refined more as we go before we actually recommend this to users.
> 
> It could be a separate PR. But the issue is that no one takes the responsibility and it falls off the radar till things go wrong. That is the reason we should insist on tests for PRs.
> 
> Also writing tests make the code more manageable into functions instead of a giant main func.
> 
> Thoughts?

I looked into writing tests and there seems to be a good example [here](https://github.com/google/go-github/blob/master/test/integration/repos_test.go), but it still requires a live Github account to make the changes on. Another option used [here](https://github.com/google/go-github/blob/887f605dd1f81715a4d4e3983e38450b29833639/github/repos_contents_test.go) is just writing endpoints to mimic github api calls, which works well for GET requests, but I'm not sure how that would work for POSTs where we want persistence after modifications. So overall testing this seems pretty difficult, but if there was a live test Github account then it would be much easier.

naveensrinivasan:
> > > > > > Without tests, it is hard to maintain things. For example that is the reason we are porting the shell script to `go` #87
> > > > > > We shouldn't merge these changes without tests and high code coverage.
> > > > > > That is one of the reasons I am blocking this change.
> > > > > 
> > > > > 
> > > > > @naveensrinivasan do you have any suggestions on how to unit test this? Since the code is mostly API calls to GH, it's difficult to write unit tests for it.
> > > > 
> > > > 
> > > > I agree unit tests could be hard for this. But we should be able to add `e2e` tests for this similar to https://github.com/ossf/scorecard/tree/main/e2e . This ensures that we haven't broken the functionality especially when there are lots of folks going to use this as a feature
> > > 
> > > 
> > > Makes sense. Since that is a big chunk of work in itself, can we tackle that as part of a separate PR? Note that, this PR is just a v0 version of the tool, something to get the framework in place. It will get refined more as we go before we actually recommend this to users.
> > 
> > 
> > It could be a separate PR. But the issue is that no one takes the responsibility and it falls off the radar till things go wrong. That is the reason we should insist on tests for PRs.
> > Also writing tests make the code more manageable into functions instead of a giant main func.
> > Thoughts?
> 
> I looked into writing tests and there seems to be a good example [here](https://github.com/google/go-github/blob/master/test/integration/repos_test.go), but it still requires a live Github account to make the changes on. Another option used [here](https://github.com/google/go-github/blob/887f605dd1f81715a4d4e3983e38450b29833639/github/repos_contents_test.go) is just writing endpoints to mimic github api calls, which works well for GET requests, but I'm not sure how that would work for POSTs where we want persistence after modifications. So overall testing this seems pretty difficult, but if there was a live test Github account then it would be much easier.

I understand! Thanks for trying and appreciate it!

multi-repo-action/scorecards.yml
laurentsimon: the scorecard yaml from scorecard repo is generally not up-to0date: it gets alerts from dependabot to be updated when we release a new action. It's cleaner to keep it in this repo

rohankh532: Something that I just thought of. The workflow needs its own PAT to run, which is usually defined as an env secret. Should we just tell the users to manually create an org-level secret that the workflow files will pick up? If we want to automate it then the script will need higher permissions. @azeemshaikh38 @laurentsimon

rohankh532: I mean the multi-repo-action script will need elevated permissions if we want it to create the secret for the org. Right now it just has basic read permissions, so I think we'd want to avoid that route. 

Then, for the workflow to run after being uploaded, the org owner will have to manually set a secrets.SCORECARD_READ_TOKEN secret on Github. 

The PAT defined in this script is just so it can read repos and make the PRs. The actual workflow will need to pull its own token from an org secret.

laurentsimon: got it. Agreed it's not ideal to have the PAT with write permission. My original idea of using workflow seems to be a bad one. Maybe we should just create a docker image and let users run it and provide their PAT on the CLI. wdut?

The PAT can be used to create the secrets (if needed for scorecard to run - this may change), and for the pull request creation. wdut?

rohankh532: That would be a good solution. I don't think the convenience is drastically different, so maybe we can save this issue for a future release and keep the current approach as a v0?

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,106 @@
...
+	"context"
+	"fmt"
+	"io/ioutil"
+
+	"github.com/google/go-github/v42/github"
+	"golang.org/x/oauth2"
+)
+
+// *****SET THESE PARAMETERS*****
+const ORG_NAME string = "ORGANIZATION NAME"
```
laurentsimon, line 13: In golang, these should be lower-case names.

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,106 @@
...
+	// Get github user client.
+	context := context.Background()
+	tokenService := oauth2.StaticTokenSource(
+		&oauth2.Token{AccessToken: PAT},
+	)
+	tokenClient := oauth2.NewClient(context, tokenService)
+	client := github.NewClient(tokenClient)
+
+	// If not provided, get all repositories under organization.
+	if len(REPO_LIST) == 0 {
```
laurentsimon, line 29: if REPO_LIST is provided by user, we should sanity-check that the org contains all these repos.

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,106 @@
...
+		err_check(err, "List Org Repos Error")
+
+		// Convert to list of repository names.
+		for _, repo := range repos {
+			REPO_LIST = append(REPO_LIST, *repo.Name)
+		}
+	}
+
+	// Get yml file into byte array.
+	workflowContent, err := ioutil.ReadFile("scorecards-analysis.yml")
```
laurentsimon, line 41: It may be useful to automatically generate the default branch and randomize the schedule, as done in https://github.com/actions/starter-workflows/blob/main/code-scanning/scorecards.yml#L6-L8 (next PR)

laurentsimon (approved):
LGTM. Try to clean up the comments by starting them with a capital latter and ending them with a full stop. Also clean global variable names following golang guidelines

rohankh532:
Ready for review. Issues to propagate to the next PR:
1. Automatically generate the default branch and randomize the schedule ([link](https://github.com/ossf/scorecard-action/pull/88#discussion_r812160517))
2. Add e2e integration tests
3. Run script as a docker image and allow users run and provide a PAT via CLI so the SCORECARD_READ_TOKEN secret can be automatically created at an org level ([link](https://github.com/ossf/scorecard-action/pull/88#discussion_r810397344))

naveensrinivasan:
@laurentsimon Is this good to be merged?

multi-repo-action/org-workflow-add.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/google/go-github/v42/github"
+	"golang.org/x/oauth2"
+)
+
+// **************************************
+// Set these parameters.
+const orgName string = "organization name"
+const pat string = "personal access token"
+
+var REPO_LIST = []string{} // Optional, leave empty to process all repos under org.
```
laurentsimon, line 17: update to RepoList

laurentsimon (approved):
Just update one of the global variable to be lower case, and we're good to merge!

multi-repo-action/org-workflow-add.go
rohankh532: fixed, sorry about that

PR #90, opened 2022-02-05
:seedling: Golang CI for clean code.

naveensrinivasan:
- Included golangci for clean code.
- Included `go` for codeql
- created a new separate action for the golangci - to keep the action file simple and also it enables them to run cocurrently.
-  Include actions cache for caching workflow runs.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,150 @@
+---
+run:
+  concurrency: 6
+  deadline: 5m
+issues:
+  new-from-rev: ""
+  include:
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -0,0 +1,150 @@
+---
```
naveensrinivasan, line 1: This is a clone of https://github.com/ossf/scorecard/blob/873308016c73b99b728e4609b23c434b28f32f6a/.goreleaser.yml#L1

PR #103, opened 2022-02-15
:seedling: Final bits of porting the shell to go

naveensrinivasan:
- Final bits of porting the shell script to `go`
- Tests included for the commandline args to Scorecard.


laurentsimon (approved):
Thanks!

PR #120, opened 2022-02-24
Sign scorecard results using cosign

rohankh532:
Added function signScorecardResult to main.go that signs the scorecard results data and creates a tlog entry in Rekor. Also added a test to main_test.go that queries Rekor and verifies that a tlog entry is actually being created.

Feature #133


azeemshaikh38:
LGTM. Cosign adds some massive amounts of dependencies :P

main.go
```diff
@@ -400,3 +410,24 @@ func runScorecardSettings(githubEventName, scorecardPolicyFile, scorecardResults
 	}
 	return &result, nil
 }
+
+func signScorecardResult(scorecardResultsFile string) error {
+	os.Setenv("COSIGN_EXPERIMENTAL", "true")
```
laurentsimon, line 415: check error of Setenv()

main.go
```diff
@@ -400,3 +410,24 @@ func runScorecardSettings(githubEventName, scorecardPolicyFile, scorecardResults
...
+	ctx := context.Background()
+
+	// Sign the scorecard result data and generate certificate and tlog entry in Rekor.
+	keyOpts := sign.KeyOpts{
+		FulcioURL:    "https://fulcio.sigstore.dev",
```
laurentsimon, line 420: maybe declare these as global variables at the top?

main.go
```diff
@@ -400,3 +410,24 @@ func runScorecardSettings(githubEventName, scorecardPolicyFile, scorecardResults
...
+		RekorURL:     "https://rekor.sigstore.dev",
+		OIDCIssuer:   options.DefaultOIDCIssuerURL,
+		OIDCClientID: "sigstore",
+	}
+	regOpts := options.RegistryOptions{}
+
+	_, err := sign.SignBlobCmd(ctx, keyOpts, regOpts, scorecardResultsFile, true, "", "", time.Minute)
+	if err != nil {
```
laurentsimon, line 428: I'm curious: Where is this function defined? I only managed to find but the signature is different https://github.com/sigstore/cosign/blob/main/cmd/cosign/cli/sign/sign_blob.go#L56 :/

main.go
```diff
@@ -400,3 +410,24 @@ func runScorecardSettings(githubEventName, scorecardPolicyFile, scorecardResults
...
+	// Sign the scorecard result data and generate certificate and tlog entry in Rekor.
+	keyOpts := sign.KeyOpts{
+		FulcioURL:    "https://fulcio.sigstore.dev",
+		RekorURL:     "https://rekor.sigstore.dev",
+		OIDCIssuer:   options.DefaultOIDCIssuerURL,
+		OIDCClientID: "sigstore",
+	}
+	regOpts := options.RegistryOptions{}
+
+	_, err := sign.SignBlobCmd(ctx, keyOpts, regOpts, scorecardResultsFile, true, "", "", time.Minute)
```
laurentsimon, line 427: Could you add a comment next to each input param to explain what they mean? It will help readers

laurentsimon, line 427: Please also add a comment why we don't need the resulting bytes

main_test.go
```diff
@@ -713,3 +716,37 @@ func Test_runScorecardSettings(t *testing.T) {
...
+		t.Errorf("signScorecardResult() error reading scorecard result file, %v", err)
+		return
+	}
+	uuids, err := cosign.FindTLogEntriesByPayload(ctx, rekorClient, scorecardResultData)
+	if err != nil {
+		t.Errorf("signScorecardResult() error getting tlog entries, %v", err)
+		return
+	}
+
+	if len(uuids) == 0 {
```
laurentsimon, line 748: is this uuid known ahead of time? I mean, can we verify it matches the one returned by ... `signScorecardResult()`?
Not sure this is do-able, just asking in case

laurentsimon (approved):
LGTM. Let's wait for @asraa to double check too.

main_test.go
```diff
@@ -713,3 +716,37 @@ func Test_runScorecardSettings(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("signScorecardResult() error getting Rekor client, %v", err)
+		return
+	}
+	scorecardResultData, err := ioutil.ReadFile("./testdata/scorecard-results-example.sarif")
+	if err != nil {
+		t.Errorf("signScorecardResult() error reading scorecard result file, %v", err)
+		return
+	}
+	uuids, err := cosign.FindTLogEntriesByPayload(ctx, rekorClient, scorecardResultData)
```
laurentsimon, line 742: @asraa is that ok if we upload into the prod TLog each time our tests run? Or is there another dev/test TLog we shoudl use?

main_test.go
```diff
@@ -713,3 +716,37 @@ func Test_runScorecardSettings(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("signScorecardResult() error getting tlog entries, %v", err)
+		return
+	}
+
+	if len(uuids) == 0 {
+		t.Errorf("signScorecardResult() error finding signature in Rekor tlog, %v", err)
```
laurentsimon, line 749: do we need to verify that the signature is valid?

laurentsimon (approved):

laurentsimon:
> LGTM. Cosign adds some massive amounts of dependencies :P

yeah that's huge!

laurentsimon:
unit test are failing:
```
Non-interactive mode detected, using device flow.
Enter the verification code TRFN-DTGH in your browser at: https://oauth2.sigstore.dev/auth/device
```

Looks like we're using a user flow for OAuth

main.go
rohankh532: This one, I think it's from a different release
https://github.com/sigstore/cosign/blob/v1.5.2/cmd/cosign/cli/sign/sign_blob.go#L58

main_test.go
rohankh532: > do we need to verify that the signature is valid?

One way to verify is by providing the certificate, signature, and transparency log index. @azeemshaikh38 mentioned that the GitHub Action will make a POST HTTP call to an endpoint that will process this verification, and the POST data payload will only contain the commitSHA, timestamp, and scorecard result payload. 

So instead of verifying normally we will use the scorecard payload to lookup all associated tlog entries, and find the one that matches our commitSHA. This way out endpoint can prove, given only a repo commit and a "claimed" scorecard result, that the results were actually generated from a scorecard run on the specified repo. 

The test in main_test.go currently doesn't account for signatures on different commits. It just submits a payload and makes sure its signature is being logged in Rekor. I wrote a function [here](https://github.com/rohankh532/scorecard-cosign-server/blob/8ced730728fdd98cf62b551e79b6cdc2acd0d68c/main.go#L32) that expands on this, and I'm working on validating the actual workflow yml contents too.

I think this design choice is so that repo owners can prove that their displayed scorecard results are authentic. Please correct if there are any misunderstandings.

rohankh532:
> unit test are failing:
> 
> ```
> Non-interactive mode detected, using device flow.
> Enter the verification code TRFN-DTGH in your browser at: https://oauth2.sigstore.dev/auth/device
> ```
> 
> Looks like we're using a user flow for OAuth

I'm not sure how to fix this. Is there a way to automatically pull keys from env vars?

main_test.go
laurentsimon: you understand correctly. If the verifying tool lives in this repo, we can perform the verification in our unit tests as well.
I agree this can be done in a follow-up PR. I was mostly asking to be sure the unit tests are complete.

@azeemshaikh38 @rohankh532 where do you think the verification code should live? In this repo or in the wepbapp repo? Im tempted to say in this repo, since both are tightly related... but I'm not entirely sure. Wdut?

laurentsimon:
> > unit test are failing:
> > ```
> > Non-interactive mode detected, using device flow.
> > Enter the verification code TRFN-DTGH in your browser at: https://oauth2.sigstore.dev/auth/device
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Looks like we're using a user flow for OAuth
> 
> I'm not sure how to fix this. Is there a way to automatically pull keys from env vars?

What do you mean? cosign private keys? Can we use the `secrets/GITHUB_TOKEN` with permissions `id-token: write`, since that's what we'll be using in the end?

main_test.go
rohankh532: I think a standalone endpoint would be useful later on, like for badging and displaying results on scorecards.dev. So maybe the webapp repo?

laurentsimon: I think that'd work, yes. Let's see what @azeemshaikh38 thinks

main_test.go
```diff
@@ -713,3 +716,37 @@ func Test_runScorecardSettings(t *testing.T) {
...
+	}
+
+	// Verify that the signature was created and uploaded to the Rekor tlog by looking up the payload.
+	ctx := context.Background()
+	rekorClient, err := rekor.NewClient("https://rekor.sigstore.dev")
+	if err != nil {
+		t.Errorf("signScorecardResult() error getting Rekor client, %v", err)
+		return
+	}
+	scorecardResultData, err := ioutil.ReadFile("./testdata/scorecard-results-example.sarif")
```
laurentsimon, line 737: `./testdata/scorecard-results-example.sarif` is used twice, let's create a variable for it

main_test.go
laurentsimon, line 748: are you sure searching for the tlog entry via payload work? Since the test will run on each push/pull request, previous runs will have added to the log the same payload. So I suspect the test passes even if you comment out the call to `signScorecardResult()`. Can you verify? If that's the case, either match their signature (which is probabilistic if we use ECDSA). Another approach is to add randomness to the payload

azeemshaikh38: +1 to keeping the verification in the webapp repo. The webapp repo will be its own standalone API (GH Action being *one* of its clients), so keeping the verification there makes sense to me.

main_test.go
asraa: currently we should use prod/public instance -- there are plans to get a dev tlog, but not yet.

main.go
```diff
@@ -400,3 +412,30 @@ func runScorecardSettings(githubEventName, scorecardPolicyFile, scorecardResults
 	}
 	return &result, nil
 }
+
+func signScorecardResult(scorecardResultsFile string) error {
+	err := os.Setenv("COSIGN_EXPERIMENTAL", "true")
```
asraa, line 417: nit: you could inline this with the error check
`if err := ... ; err != nil { }`

asraa:
> What do you mean? cosign private keys? Can we use the secrets/GITHUB_TOKEN with permissions id-token: write, since that's what we'll be using in the end?

Yeah if you're running this unit test in CI we can pull from the environment variables. I will talk offline or add inline to the code

main.go
```diff
@@ -72,6 +77,8 @@ const (
 	scorecardPolicyFile = "./policy.yml"
 	scorecardFork       = "SCORECARD_IS_FORK"
 	sarif               = "sarif"
+	fulcioURL           = "https://fulcio.sigstore.dev"
```
asraa, line 80: maybe remove this and use
https://github.com/sigstore/cosign/blob/d2781b81e9d0bb6149aec4dd99bffee9fe197fc8/cmd/cosign/cli/options/rekor.go#L22
and 
https://github.com/sigstore/cosign/blob/d2781b81e9d0bb6149aec4dd99bffee9fe197fc8/cmd/cosign/cli/options/fulcio.go#L22

main.go
```diff
@@ -400,3 +412,30 @@ func runScorecardSettings(githubEventName, scorecardPolicyFile, scorecardResults
...
+		OIDCIssuer:   options.DefaultOIDCIssuerURL, // OIDC provider to get ID token to auth for Fulcio.
+		OIDCClientID: "sigstore",
+	}
+	regOpts := options.RegistryOptions{} // Not necessary so we leave blank.
+
+	// This command will use the provided OIDCIssuer to authenticate into Fulcio, which will generate the
+	// signing certificate on the scorecard result. This attestation is then uploaded to the Rekor transparency log.
+	// The output bytes (signature) and certificate are discarded since verification can be done with just the payload.
+	_, err = sign.SignBlobCmd(ctx, keyOpts, regOpts, scorecardResultsFile, true, "", "", time.Minute)
+	if err != nil {
```
asraa, line 436: nit: you could also check error on the same line since the other rval isn't used

main_test.go
asraa: agreed, the test will always pass. 

> If that's the case, either match their signature (which is probabilistic if we use ECDSA).

We can return the signature from the `SignBlobCmd` and then in `signScorecardResult()`. You can iterate through the UUIDs and extract their signatures (see this similar [func](https://github.com/sigstore/cosign/blob/18d2ce0b458018951f7356db911467a427a8dffe/cmd/cosign/cli/verify/verify_blob.go#L345) that extracts the certificate from the hashedrekord payload, you want the signature payload), and check if one matches.

laurentsimon: I wonder whether it's better to randomize the payload for this use case. My worry is that the search is going to grow O(n^2) with n being he number of runs. This is not a problem for genuine releases because projects release a limited number of releases in their lifetime (although it'd still be nice for rekor to have a "commit has index" because it'd make things faster)

I first thought the commit hash index could also be a mitigation against DoS since anyone can add random entries with the same payload... but I guess anyone can add entries for the same hash too :-) It'd be nice to mitigate these sort of attack by having rekor aware of GitHub, and verify that the entry was generated from GitHub workflow before adding the entry to the log. I think we could add an additional flag in the cert, like "origin", but I suspect it's already possible to infer this given the cert? @asraa wdut?

asraa: > . My worry is that the search is going to grow O(n) with n being the number of runs

You're right -- each test invocation is going to cause an additional UUID to search. Let's add a nonce to the payload.

> and verify that the entry was generated from GitHub workflow before adding the entry to the log.

Why should rekor prevent someone from uploading a verified entry? Or are you suggesting that if rekor sees a github workflow certificate in an entry, that it should verify that the upload was happening inside a workflow? I think I disagree -- rekor already verifies that the signature is valid on upload, so this is "probably" enough to mitigate a DoS -- you have to go through the process of signing before you can stick something in the log. It also de-dupes entries, so you can't spam rekor unless you went through the process of signing inside the workflow.

main_test.go
asraa, line 742: This is an FYI for later: `FindTLogEntriesByPayload` uses the search index on top of rekor to find the payload. it does NOT verify the UUIDs are actually in the log! Probably fine for a unit test, but the API service validating will need to do additional verification to check if the UUID is part of the tlog.

(see https://github.com/sigstore/cosign/blob/18d2ce0b458018951f7356db911467a427a8dffe/pkg/cosign/tlog.go#L225)

main_test.go
rohankh532: I think adding some randomness to the payload is a good idea.

> Or are you suggesting that if rekor sees a github workflow certificate in an entry, that it should verify that the upload was happening inside a workflow?

Isn't it necessary to prove that the upload came from the scorecard workflow? What if someone tampered with the results and then used the same process to sign and upload them. Then, when verifying, we will lookup the payload and see a corresponding tlog entry.

main_test.go
rohankh532: `verifyTlogEntry` is private so I can't import it. Is there any workaround?

main_test.go
laurentsimon: probably unrelated question: can rekor be queried by signature instead of payload? Note that there are also re-builders who may sign reproducible builds, so the payload may get duplicated: is that correct? What do you mean by `de-dupes entries`?

@asraa IIUC, rekor verifies the signature chains up to Fulcio root before allowing the entry to be added? (so it forces a user to perform at least the OIDC flow). How about non-OIDC flows?

who generates the public key, the client or Fulcio?

naveensrinivasan:
> LGTM. Cosign adds some massive amounts of dependencies :P

This is a concern that has been brought in cosign slack channel. Package reorganization is required.

main.go
```diff
@@ -400,3 +410,28 @@ func runScorecardSettings(githubEventName, scorecardPolicyFile, scorecardResults
...
+		RekorURL:     options.DefaultRekorURL,      // Transparency log.
+		OIDCIssuer:   options.DefaultOIDCIssuerURL, // OIDC provider to get ID token to auth for Fulcio.
+		OIDCClientID: "sigstore",
+	}
+	regOpts := options.RegistryOptions{} // Not necessary so we leave blank.
+
+	// This command will use the provided OIDCIssuer to authenticate into Fulcio, which will generate the
+	// signing certificate on the scorecard result. This attestation is then uploaded to the Rekor transparency log.
+	// The output bytes (signature) and certificate are discarded since verification can be done with just the payload.
+	if _, err := sign.SignBlobCmd(ctx, keyOpts, regOpts, scorecardResultsFile, true, "", "", time.Minute); err != nil {
```
naveensrinivasan, line 432: Signing this should be optional to the users. Some users might not want to do this, I would recommend we have a flag to enable/disable this setting.

naveensrinivasan (changes requested):
After some thought, I think this should be within Scorecard and not part of Scorecard action. Similar to my comment we should provide an option to the end-users either use public rekor to store the artifacts or not.

main.go
laurentsimon: when users set the option `publish_results: true` in their workflow, we'll publish the results and sign them (if needed) for the badge. Otherwise this will be skipped. 

Does this resolve the concern?

laurentsimon:
> After some thought, I think this should be within Scorecard and not part of Scorecard action. Similar to my comment we should provide an option to the end-users either use public rekor to store the artifacts or not.

I suppose we can do that next. But it's not clear whether allowing users to sign-and-push to rekor would be helpful. The whole key management problem becomes an issue during verification, does it not? The signature within GH workflow allows us to mitigate this problem, at least if we provide the tooling to verify it out of the box. Wdut?

main.go
naveensrinivasan: Yes, Thanks. We would  have to document in the action that publishing results also pushes it to rekor.

naveensrinivasan:
> > After some thought, I think this should be within Scorecard and not part of Scorecard action. Similar to my comment we should provide an option to the end-users either use public rekor to store the artifacts or not.
> 
> It's not clear whether allowing users to sign-and-push to rekor within scorecard would be helpful. The whole key management problem becomes an issue during verification. The signature within GH workflow allows us to mitigate this problem, at least if we provide the tooling to verify it out of the box. Wdut?

Agreed, I completely forgot about that. Thanks for reminding me about the key issue. 

We could use the cron to sign and push to rekor. If I am not wrong there is an existing issue in Scorecard.

naveensrinivasan (approved):
Thanks, LGTM except other than documenting in the Action as a comment that when publish_results is true we use the public rekor to store the signed artifacts.

azeemshaikh38:
> After some thought, I think this should be within Scorecard and not part of Scorecard action. Similar to my comment we should provide an option to the end-users either use public rekor to store the artifacts or not.

What is the value-add by moving this logic to Scorecard? We should allow users to sign their Scorecard results (if they want to) separately from Scorecard and not have this baked into the tool itself. 

Having this here makes sense because we are signing results by a *trusted* entity, i.e  a GitHub workflow. Also, regarding storing in Rekor:

a) We are only saving the signature not the artifact, so it might be ok to do it for every run.
b) If storing signatures is also a concern, we can use the `publish_results` bool to make the distinction.

azeemshaikh38:
> 

Ah, didn't realize this is already resolved.

azeemshaikh38:
> > After some thought, I think this should be within Scorecard and not part of Scorecard action. Similar to my comment we should provide an option to the end-users either use public rekor to store the artifacts or not.
> 
> What is the value-add by moving this logic to Scorecard? We should allow users to sign their Scorecard results (if they want to) separately from Scorecard and not have this baked into the tool itself.
> 
> Having this here makes sense because we are signing results by a _trusted_ entity, i.e a GitHub workflow. Also, regarding storing in Rekor:
> 
> a) We are only saving the signature not the artifact, so it might be ok to do it for every run. b) If storing signatures is also a concern, we can use the `publish_results` bool to make the distinction.

Ah, didn't realize this is already resolved.

main.go
```diff
@@ -124,6 +129,11 @@ func main() {
 		panic(err)
 	}
 
+	err = signScorecardResult(scorecardResultsFile)
+	if err != nil {
+		panic(err)
```
azeemshaikh38, line 134: Let's not `panic` here. Upload to Rekor may fail and we don't want that to show up as failures to users. Instead let's do this - log this error and quietly return. In a future PR when you add the POST call to the API, make the error a part of the POST payload. On the API side, we can log these errors so that we can monitor these errors through there.

@asraa fyi who brought up this concern.

azeemshaikh38 (approved):

main.go
```diff
@@ -129,9 +129,11 @@ func main() {
 		panic(err)
 	}
 
-	err = signScorecardResult(scorecardResultsFile)
-	if err != nil {
-		panic(err)
+	if scorecardPublishResults == "true" {
+		err = signScorecardResult(scorecardResultsFile)
+		if err != nil {
+			return
```
naveensrinivasan, line 135: Can we please log the error? This would help troubleshoot issues.

naveensrinivasan (commented):
Thanks for taking care of the publish_results flag.

main.go
```diff
@@ -132,6 +132,7 @@ func main() {
 	if scorecardPublishResults == "true" {
 		err = signScorecardResult(scorecardResultsFile)
 		if err != nil {
+			fmt.Printf("Error signing scorecard result: %v", err)
```
naveensrinivasan, line 135: Thanks, could you instead please write to stderr `fmt.Fprintf(os.Stderr, "Error signing scorecard result: %v", err)`?

justaugustus (commented):
EXPLICIT HOLD on this PR until https://github.com/ossf/scorecard-action/pull/122 merges.
The rebase is going to be really painful otherwise.

justaugustus:
@rohankh532 -- Let me know when you're ready for a review here! :)

Dockerfile
```diff
@@ -21,23 +21,29 @@
...
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build
+
 # TODO: use distroless.
+# FROM gcr.io/distroless/base:nonroot@sha256:02f667185ccf78dbaaf79376b6904aea6d832638e1314387c2c2932f217ac5cb
 FROM debian:11.2-slim@sha256:40f90ead34f61f38c69fd604878c73c75ec944055328b9883f3b643c1b8886f0
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     jq ca-certificates curl
```
laurentsimon, line 43: curl no longer needed... maybe keep it for debugging?

signing/signing.go
```diff
@@ -0,0 +1,105 @@
...
+	"net/http"
+	"os"
+
+	"github.com/ossf/scorecard-action/entrypoint"
+	"github.com/ossf/scorecard-action/options"
+	sigOpts "github.com/sigstore/cosign/cmd/cosign/cli/options"
+	"github.com/sigstore/cosign/cmd/cosign/cli/sign"
+)
+
+func SignScorecardResult(scorecardResultsFile string) error {
```
laurentsimon, line 17: does this need to be public or can it be kept private?

signing/signing.go
```diff
@@ -0,0 +1,105 @@
...
+// Calls scorecard-api to process & upload signed scorecard results.
+// TODO: not sure how to test this because it requires running the entire scorecard action.
+func ProcessSignature() error {
+	// Get sarif output from file.
+	sarifPayload, err := ioutil.ReadFile(os.Getenv(options.EnvInputResultsFile))
+	if err != nil {
+		return fmt.Errorf("reading scorecard sarif results from file: %v", err)
+	}
+
+	// Change output settings to json and run scorecard again.
```
laurentsimon, line 52: add a note that this will be changed later to run scorecard only once, and generate multiple format. Maybe create a tracking issue for it.

signing/signing.go
```diff
@@ -0,0 +1,105 @@
...
+
+	payloadBytes, err := json.Marshal(resultsPayload)
+	if err != nil {
+		return fmt.Errorf("reading scorecard json results from file: %v", err)
+	}
+
+	// Call scorecard-webapp-api to process and upload signature.
+	url := "https://api.securityscorecards.dev/verify"
+	req, _ := http.NewRequest("POST", url, bytes.NewBuffer(payloadBytes))
+	req.Header.Set("Repository", os.Getenv(options.EnvGithubRepository))
```
laurentsimon, line 90: those are HTTP headers? Shall we set them as `X-Repository` and `X-Branch` since they are custom headers?

signing/signing_test.go
```diff
@@ -0,0 +1,59 @@
...
+		t.Errorf("signScorecardResult() error reading scorecard result file, %v", err)
+		return
+	}
+	uuids, err := cosign.FindTLogEntriesByPayload(ctx, rekorClient, scorecardResultData)
+	if err != nil {
+		t.Errorf("signScorecardResult() error getting tlog entries, %v", err)
+		return
+	}
+
+	if len(uuids) == 0 {
```
laurentsimon, line 55: I think you can even be more specific and use `if len(uuids) != 1`

laurentsimon (approved):

signing/signing.go
rohankh532: changed to private

rohankh532: ended up changing it back to public because it made more sense to call this from main.go than from ProcessSignature in signing.go

signing/signing.go
```diff
@@ -0,0 +1,106 @@
...
+	jsonPayload, err := ioutil.ReadFile(os.Getenv(options.EnvInputResultsFile))
+	if err != nil {
+		return nil, fmt.Errorf("reading scorecard json results from file: %v", err)
+	}
+
+	return jsonPayload, nil
+}
+
+// Calls scorecard-api to process & upload signed scorecard results.
+func ProcessSignature(sarifPayload []byte, jsonPayload []byte, repoName string, repoRef string) error {
```
asraa, line 71: small suggestion! with go you can combine type params if the subsequent ones are the same.. e.g.:
```
ProcessSignature(sarifPayload, jsonPayload []byte, repoName, repoRef string)
```

signing/signing.go
```diff
@@ -0,0 +1,106 @@
...
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("executing scorecard-api call: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != 200 {
+		bodyBytes, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("http response error: %v", string(bodyBytes))
```
asraa, line 102: might be worth adding the `resp.Status` and `resp.StatusCode` in the error string -- in case there's not body in the not OK HTTP response, it will be hard to debug whether this was a 5xx or 4xx or other type of error.

asraa (approved):
Very nice work!! LGTM

Issue #107, opened 2022-02-17
Use Scorecard's Go APIs


laurentsimon:
The entrypoint.sh uses env variables because it was written s a bash script.
The current main.go follows the same design.

We can make the code simpler and more go-native by getting rid of unnecessary env variables.

Env variables that are necessary before running scorecard are https://github.com/ossf/scorecard-action/blob/main/entrypoint.sh#L26-L28 - basically the ones required via command line.

Other env variables, such as https://github.com/ossf/scorecard-action/blob/main/entrypoint.sh#L29-L34 and https://github.com/ossf/scorecard-action/blob/main/entrypoint.sh#L51-L53 can be safely replaced by native go variables.

@naveensrinivasan is this something you have bandwidth for?

cc @azeemshaikh38 @rohankh532

azeemshaikh38:
+1. In fact, I think we should Golang-ify the code more. Especially, we should move away from running Scorecard as a bash script and run it as a library instead.

naveensrinivasan:
>@naveensrinivasan is this something you have bandwidth for?

I can do this after I am done with the things I am working on. I think this can wait.

azeemshaikh38:
> @rohankh532 will work on using API instead of CLI as part of the crowdsourcing the results. Is that still the plan?

@rohankh532 has agreed to work on this. Thanks Rohan!

laurentsimon:
This is done. Closing.

azeemshaikh38:
Re-opening. I thought the plan was to move to using Scorecard's Go APIs instead of running it in a bash process - https://github.com/ossf/scorecard-action/issues/107#issuecomment-1043193349?

laurentsimon:
you're right. I though we could create a different issue for this since "remove env variable" is effectively done. Fine to keep it here, though.

azeemshaikh38:
Makes sense. I updated the title.

naveensrinivasan:
@rohankh532 Are you planning to finish this up? I am planning to write some tests which will change based on this. I wanted to understand so that I can plan based on that.

rohankh532:
> @rohankh532 Are you planning to finish this up? I am planning to write some tests which will change based on this. I wanted to understand so that I can plan based on that.

Yeah I can. Are there any docs on Scorecard's Go APIs?

naveensrinivasan:
Here is the CLI code that parses the command line and invokes the API https://github.com/ossf/scorecard/blob/v4.1.0/cmd/root.go

[...]

PR #122, opened 2022-02-27
Create a Golang-based entrypoint for scorecard-action (1/n)

justaugustus:
Part of https://github.com/ossf/scorecard-action/issues/107.


laurentsimon:
cc @rohankh532

naveensrinivasan (commented):
Looks good. Thank you!, I didn't approve it because I would like @laurentsimon also to review it.

options/options.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := opts.Initialize(); err != nil {
+		return opts, fmt.Errorf(
+			"initializing scorecard-action options: %w",
+			err,
+		)
+	}
+
+	// TODO(options): Move this set-or-default logic to its own function.
+	opts.ScorecardOpts.EnableSarif = true
+	if opts.ScorecardOpts.Format == formatSarif {
```
azeemshaikh38, line 96: Nit: use a single `if` instead of nested for better readability.

github/github.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/ossf/scorecard/v4/log"
+)
+
+const (
+	baseRepoURL = "https://api.github.com/repos/"
+)
+
+// RepoInfo is a struct for repository information.
+type RepoInfo struct {
+	Repo      repo `json:"repository"`
```
azeemshaikh38, line 36: Curious - since `repo` struct is not exposed how does an external package access this nested struct?

azeemshaikh38 (commented):
This is awesome @justaugustus, thanks! Same comment as @naveensrinivasan, will wait for @laurentsimon review.

github/github.go
justaugustus: > Curious - since `repo` struct is not exposed how does an external package access this nested struct?

Great question! Really, you want to ideally discourage direct usage.
Constructing a client using `NewClient(ctx)` is going to provide a consumer with a properly populated `RepoInfo` struct.

Thinking about the context in which we're running this code, we've got two use cases:
- "production" -- this is running within the environment of a GitHub Action, where the event path and thereby the repository information will be present to unmarshal into `RepoInfo`
- testing -- this is where we need to ensure the developer environment/tests are scaffolded in a way that mimics the production env (`testdata/` / fixtures)

In both situations, I _think_ it's not strictly required to have these fields exported, but I'm open to it, if needed.

options/options.go
justaugustus: > Nit: use a single `if` instead of nested for better readability.

@azeemshaikh38 -- I wrestled with this a little...

- `scorecard` currently defaults to `FormatDefault` (which I believe is the table writer)
- using `scorecard.options.New()` will get you the default values of the options, but can be overridden during initialization with env vars, like `SCORECARD_RESULTS_FORMAT`
- `scorecard-action` really wants a default format of `options.FormatSarif`, which means:
- you also need to ensure that `ENABLE_SARIF` is set in the enviroment (because the SARIF option is feature-flagged)
- only when SARIF format is chosen AND `ENABLE_SARIF` is set will `PolicyFile` be available as a CLI flag, which means:
- you should only expose the option to set `PolicyFile` when you know SARIF is enabled

I asked in the channel, but I'll ask here too... do we want not bother exposing the format here altogether and just default to SARIF for `scorecard-action`?

justaugustus:
I want to call out the test failures here:

Unit tests (https://github.com/ossf/scorecard-action/runs/5426123389?check_suite_focus=true):

```console
Run go test -covermode=atomic  -coverprofile=unit-coverage.out ./...
  go test -covermode=atomic  -coverprofile=unit-coverage.out ./...
  shell: /usr/bin/bash -e {0}
  env:
    GOROOT: /opt/hostedtoolcache/go/1.17.7/x64
?   	github.com/ossf/scorecard-action	[no test files]
ok  	github.com/ossf/scorecard-action/entrypoint	0.009s	coverage: 0.0% of statements [no tests to run]
?   	github.com/ossf/scorecard-action/github	[no test files]
--- FAIL: TestNew (0.00s)
    --- FAIL: TestNew/SuccessFormatSARIF (0.00s)
        options_test.go:107: New() error = validating scorecard options: some options could not be validated: [commit option is not supported yet], wantErr false
    --- FAIL: TestNew/SuccessFormatJSON (0.00s)
        options_test.go:107: New() error = validating scorecard options: some options could not be validated: [commit option is not supported yet], wantErr false
FAIL
coverage: 43.9% of statements
FAIL	github.com/ossf/scorecard-action/options	0.012s
FAIL
Error: Process completed with exit code 1.
```

golangci-lint (https://github.com/ossf/scorecard-action/runs/5426123392?check_suite_focus=true):

```console
Run golangci/golangci-lint-action@b517f99ae23d86ecc4c0dec08dcf48d2336abc29
prepare environment
run golangci-lint
  Running [/home/runner/golangci-lint-1.44.0-linux-amd64/golangci-lint run --out-format=github-actions --new-from-patch=/tmp/tmp-1656-d52rcyaL5U29/pull.patch --new=false --new-from-rev=] in [] ...
  Error: File is not `gci`-ed (gci)
  Error: File is not `gci`-ed (gci)
  Error: File is not `gci`-ed (gci)
  
  Error: issues found
  Ran golangci-lint in 3505ms
```

VS what happens locally...

```console
❯ time go test -v -covermode=atomic  -coverprofile=unit-coverage.out ./...
?   	github.com/ossf/scorecard-action	[no test files]
testing: warning: no tests to run
PASS
coverage: 0.0% of statements
ok  	github.com/ossf/scorecard-action/entrypoint	0.354s	coverage: 0.0% of statements [no tests to run]
?   	github.com/ossf/scorecard-action/github	[no test files]
=== RUN   TestNew
=== RUN   TestNew/SuccessFormatSARIF
=== RUN   TestNew/SuccessFormatJSON
--- PASS: TestNew (0.00s)
    --- PASS: TestNew/SuccessFormatSARIF (0.00s)
    --- PASS: TestNew/SuccessFormatJSON (0.00s)
=== RUN   TestInitialize
=== RUN   TestInitialize/Success
=== RUN   TestInitialize/FailureNoFieldsSet
=== RUN   TestInitialize/FailureBadEventPath
=== RUN   TestInitialize/FailureBadEventData
--- PASS: TestInitialize (0.00s)
    --- PASS: TestInitialize/Success (0.00s)
    --- PASS: TestInitialize/FailureNoFieldsSet (0.00s)
    --- PASS: TestInitialize/FailureBadEventPath (0.00s)
    --- PASS: TestInitialize/FailureBadEventData (0.00s)
PASS
coverage: 54.4% of statements
ok  	github.com/ossf/scorecard-action/options	0.226s	coverage: 54.4% of statements
go test -v -covermode=atomic -coverprofile=unit-coverage.out ./...  1.86s user 1.12s system 165% cpu 1.798 total
```

```console
❯ time golangci-lint run -c .golangci.yml
golangci-lint run -c .golangci.yml  1.33s user 1.01s system 210% cpu 1.116 total
```

Something is up with CI tests; will check it out.

azeemshaikh38:
> Something is up with CI tests; will check it out.

A wild guess, the `golangci-lint` that is being downloaded by CI (through `make install`) is probably pulling some version that's broken (or atleast different than your local instance).

justaugustus:
> A wild guess, the `golangci-lint` that is being downloaded by CI (through `make install`) is probably pulling some version that's broken (or atleast different than your local instance).

Yep! I'm checking out https://github.com/nektos/act locally to see if I can mimic the GitHub Action env.

options/options.go
azeemshaikh38: > do we want not bother exposing the format here altogether and just default to SARIF for `scorecard-action`?

I think having SARIF as the default format in `scorecard-action` is fine, but users may want to dump out a JSON instead of SARIF?

Even though I can't think of a usecase where someone is using JSON instead of SARIF, the work involved in un-exposing the format option (we'll need to make sure current users aren't broken) is far greater than the benefit it gets us.

github/github.go
azeemshaikh38: Thanks for the explanation. Leaving it as-is LGTM.

justaugustus:
> Nice! That's a pretty cool tool. @laurentsimon might help in testing `scorecard-action` locally.

Exactly what I was thinking!

options/options.go
justaugustus: > I think having SARIF as the default format in `scorecard-action` is fine, but users may want to dump out a JSON instead of SARIF?
> 
> Even though I can't think of a usecase where someone is using JSON instead of SARIF, the work involved in un-exposing the format option (we'll need to make sure current users aren't broken) is far greater than the benefit it gets us.

Yep, this would be dramatically simpler without the env vars, etc., but I want to ensure backwards-compat, so I agree that it's worth keeping both.

justaugustus:
> > Something is up with CI tests; will check it out.
> 
> A wild guess, the `golangci-lint` that is being downloaded by CI (through `make install`) is probably pulling some version that's broken (or atleast different than your local instance).

@azeemshaikh38 @naveensrinivasan -- Can you take a peek at https://github.com/ossf/scorecard-action/pull/126? That PR fixes the lint config.

justaugustus:
@laurentsimon @azeemshaikh38 @naveensrinivasan -- I think this is a good stopping point. There's a bunch more to do, but all of the tests are passing and I'd to unblock the work happening in https://github.com/ossf/scorecard-action/pull/120 (cc: @rohankh532).

github/github.go
```diff
@@ -0,0 +1,161 @@
...
+
+	repoURL, err := baseURL.Parse(repoName)
+	if err != nil {
+		return r, fmt.Errorf("parsing repo endpoint: %w", err)
+	}
+
+	method := "GET"
+	req, err := http.NewRequestWithContext(
+		c.ctx,
+		method,
```
azeemshaikh38, line 131: Can use `http.MethodGet` here instead.

azeemshaikh38 (approved):
LGTM, thanks!

PR #136, opened 2022-03-09
Fixups for golang-based entrypoint

rohankh532:
1. ~~(Dockerfile) Fixed the dockerfile so that it uses the new golang entrypoint instead of entrypoint.sh~~ update later
2. (Makefile) Created a makefile for building scorecard-action
3. ~~(Dockerfile, options.go) Fixed the defaultScorecardPolicyFile not being copied~~ update later
4. (entrypoint.go) Changed resultsFilePath so that it is under the GithubWorkspace dir to fix file permission errors.
5. (options.go) Properly pull & set EnvInputResultsFormat, EnvInputResultsFile, and EnvGithubAuthToken env vars
6. (options_test.go) Set EnvInputResultsFormat and EnvInputResultsFile before calling options.New() to see if it properly picks up env vars.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,23 +21,28 @@
 #           -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN \
 #           -e GITHUB_REPOSITORY="ossf/scorecard" \
 #           laurentsimon/scorecard-action:latest
-FROM gcr.io/openssf/scorecard:v4.1.0@sha256:a1e9bb4a0976e800e977c986522b0e1c4e0466601642a84470ec1458b9fa6006 as base
 
-# Build our image and update the root certs.
-# TODO: use distroless.
+#v1.17 go
+FROM golang@sha256:bd9823cdad5700fb4abe983854488749421d5b4fc84154c30dae474100468b85 AS base
+WORKDIR /src
+ENV CGO_ENABLED=0
+COPY go.* ./
+RUN go mod download
+COPY . ./
+
+FROM base AS build
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build-scorecard-action
+
+# FROM gcr.io/distroless/base:nonroot@sha256:02f667185ccf78dbaaf79376b6904aea6d832638e1314387c2c2932f217ac5cb
 FROM debian:11.2-slim@sha256:d5cd7e54530a8523168473a2dcc30215f2c863bfa71e09f77f58a085c419155b
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     jq ca-certificates curl
-
-# Copy the scorecard binary from the official scorecard image.
-COPY --from=base /scorecard /scorecard
+COPY --from=build /src/scorecard-action /
 
 # Copy a test policy for local testing.
 COPY policies/template.yml  /policy.yml
 
-# Our entry point.
-# Note: the file is executable in the repo
-# and permission carry over to the image.
-COPY entrypoint.sh /entrypoint.sh
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT [ "/scorecard-action" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,29 +21,23 @@
 #           -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN \
 #           -e GITHUB_REPOSITORY="ossf/scorecard" \
 #           laurentsimon/scorecard-action:latest
+FROM gcr.io/openssf/scorecard:v4.1.0@sha256:a1e9bb4a0976e800e977c986522b0e1c4e0466601642a84470ec1458b9fa6006 as base
 
-#v1.17 go
-FROM golang@sha256:bd9823cdad5700fb4abe983854488749421d5b4fc84154c30dae474100468b85 AS base
-WORKDIR /src
-ENV CGO_ENABLED=0
-COPY go.* ./
-RUN go mod download
-COPY . ./
-
-FROM base AS build
-ARG TARGETOS
-ARG TARGETARCH
-RUN CGO_ENABLED=0 make build-scorecard-action
-
+# Build our image and update the root certs.
 # TODO: use distroless.
-# FROM gcr.io/distroless/base:nonroot@sha256:02f667185ccf78dbaaf79376b6904aea6d832638e1314387c2c2932f217ac5cb
 FROM debian:11.2-slim@sha256:d5cd7e54530a8523168473a2dcc30215f2c863bfa71e09f77f58a085c419155b
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
     jq ca-certificates curl
-COPY --from=build /src/scorecard-action /
+
+# Copy the scorecard binary from the official scorecard image.
+COPY --from=base /scorecard /scorecard
 
 # Copy a test policy for local testing.
 COPY policies/template.yml  /policy.yml
 
-ENTRYPOINT [ "/scorecard-action" ]
+# Our entry point.
+# Note: the file is executable in the repo
+# and permission carry over to the image.
+COPY entrypoint.sh /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,6 +35,7 @@ ARG TARGETOS
 ARG TARGETARCH
 RUN CGO_ENABLED=0 make build-scorecard-action
 
+# TODO: use distroless.
 # FROM gcr.io/distroless/base:nonroot@sha256:02f667185ccf78dbaaf79376b6904aea6d832638e1314387c2c2932f217ac5cb
 FROM debian:11.2-slim@sha256:d5cd7e54530a8523168473a2dcc30215f2c863bfa71e09f77f58a085c419155b
 RUN apt-get update && \
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 LDFLAGS=-w -extldflags
 
-build-scorecard-action: ## Runs go build on repo
+build: ## Runs go build on repo
 	# Run go build and generate scorecard executable
 	CGO_ENABLED=0 go build -o scorecard-action -trimpath -a -tags netgo -ldflags '$(LDFLAGS)'
+	
--- a/Makefile
+++ b/Makefile
@@ -3,4 +3,3 @@ LDFLAGS=-w -extldflags
 build: ## Runs go build on repo
 	# Run go build and generate scorecard executable
 	CGO_ENABLED=0 go build -o scorecard-action -trimpath -a -tags netgo -ldflags '$(LDFLAGS)'
-	
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+LDFLAGS=-w -extldflags
+
+build-scorecard-action: ## Runs go build on repo
+	# Run go build and generate scorecard executable
+	CGO_ENABLED=0 go build -o scorecard-action -trimpath -a -tags netgo -ldflags '$(LDFLAGS)'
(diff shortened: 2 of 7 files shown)


Dockerfile
```diff
@@ -21,23 +21,29 @@
 #           -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN \
```
azeemshaikh38, line 21: Let's not update the Dockerfile in this PR. It will roll out Golang based scorecard-action to end users (whenever we create a new release or for anyone brave enough to be using `@main` ref).

makefile
```diff
@@ -0,0 +1,3 @@
+build-scorecard-action: ## Runs go build on repo
```
azeemshaikh38, line 1: `makefile` naming convention is `Makefile`.

makefile
```diff
@@ -0,0 +1,3 @@
...
+	# Run go build and generate scorecard executable
+	CGO_ENABLED=0 go build -trimpath -a -tags netgo -ldflags '$(LDFLAGS)'
```
azeemshaikh38, line 3: `LDFLAGS` is undefined. Maybe add an LDFLAGS var like [here](https://github.com/ossf/scorecard-webapp/blob/main/Makefile#L1).

Also, explicitly call out the binary name this command will output using `-o scorecard-action`

options/options_test.go
```diff
@@ -128,6 +128,17 @@ func TestNew(t *testing.T) {
 				}
 			}
 
+			if tt.repo != "" {
+				os.Setenv(EnvInputResultsFormat, tt.resultsFormat)
+			}
+
+			_, resultsfileEnvExists := os.LookupEnv(EnvInputResultsFile)
```
azeemshaikh38, line 135: Any reason to assert whether this is set or not before setting this env var?

azeemshaikh38 (commented):
LGTM overall. Will let @justaugustus and @laurentsimon also take a look.

Dockerfile
```diff
@@ -21,23 +21,29 @@
...
 #           -e GITHUB_REPOSITORY="ossf/scorecard" \
 #           laurentsimon/scorecard-action:latest
-FROM gcr.io/openssf/scorecard:v4.1.0@sha256:a1e9bb4a0976e800e977c986522b0e1c4e0466601642a84470ec1458b9fa6006 as base
 
-# Build our image and update the root certs.
+#v1.17 go
+FROM golang@sha256:bd9823cdad5700fb4abe983854488749421d5b4fc84154c30dae474100468b85 AS base
```
justaugustus, line 26: Reference the version number as well:

`golang:<tag>@<digest>`

options/options_test.go
rohankh532: Removed, this was a remnant from before

Dockerfile
```diff
@@ -21,23 +21,29 @@
...
+WORKDIR /src
+ENV CGO_ENABLED=0
+COPY go.* ./
+RUN go mod download
+COPY . ./
+
+FROM base AS build
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build-scorecard-action
```
justaugustus, line 36: While true in this instance, don't assume that the Makefile is going to exist/function as expected in the build env.
Prefer `go build ...`

justaugustus (changes requested):
Left a few initial pass comments!

entrypoint/entrypoint.go
```diff
@@ -60,7 +60,8 @@ func New() (*cobra.Command, error) {
 
 		if scOpts.ResultsFile != "" {
 			var err error
-			out, err = os.Create(scOpts.ResultsFile)
+			resultsFilePath := fmt.Sprintf("%v/%v", opts.GithubWorkspace, scOpts.ResultsFile)
+			out, err = os.Create(resultsFilePath)
```
laurentsimon, line 64: update the `fmt.Errorf` call below to use `resultsFilePath`

options/options.go
```diff
@@ -107,6 +108,15 @@ func New() (*Options, error) {
 	}
 
 	opts.SetPublishResults()
+
+	opts.ScorecardOpts.ResultsFile = os.Getenv(EnvInputResultsFile)
+
+	// Set PAT if not provided (locally).
+	_, repoTokenExists := os.LookupEnv(EnvGithubAuthToken)
```
laurentsimon, line 115: in which scenario is the EnvGithubAuthToken set and not set? It's set by GitHub, so we don't control it, right?

Note that we could set the default PAT to be the GH token in this file https://github.com/ossf/scorecard-action/blob/main/action.yaml#L30

laurentsimon (approved):
LGTM overall, added a few comments to clarify my understanding. Thanks!

options/options.go
rohankh532: Looks like there was a redundant env var (GITHUB_AUTH_TOKEN) being set. fixed this so it directly uses the input.

options/options.go
```diff
@@ -91,6 +91,7 @@ func New() (*Options, error) {
 	}
 
 	// TODO(options): Move this set-or-default logic to its own function.
+	opts.ScorecardOpts.Format = os.Getenv(EnvInputResultsFormat)
```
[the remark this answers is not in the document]
rohankh532: The struct fields for [ScorecardOpts](https://github.com/ossf/scorecard/blob/4b9f0389c6f6/options/options.go#L30) are named differently (ex. "SCORECARD_RESULTS_FORMAT" instead of "INPUT_RESULTS_FORMAT"), so env.Parse won't work here. Is there a workaround?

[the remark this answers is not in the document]
laurentsimon: I think it's fine to update the names to `INPUT_RESULTS_FORMAT`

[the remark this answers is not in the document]
azeemshaikh38: Should we update `options/env.go` to contain a struct instead of a list of `consts`? So basically,

```
// options/env.go

type Env struct {
  EnableSarif `env:ENABLE_SARIF`
  AuthToken `env:INPUT_REPO_TOKEN`
 ...
}

// options/options.go

env.Parse(Env)
```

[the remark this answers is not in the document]
rohankh532: > I think it's fine to update the names to `INPUT_RESULTS_FORMAT`

I don't think I can edit that file since its importing from a previous commit.

> Should we update `options/env.go` to contain a struct instead of a list of `consts`? So basically,

Yeah this would make the code a lot more clean. Right now there are a lot of duplicate vars for the env vars.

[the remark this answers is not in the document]
rohankh532: I updated options/env.go to be a struct so it reduced the amount of OS calls in options. Also tested and verified that the workflow runs properly.

naveensrinivasan (approved):

Makefile
```diff
@@ -0,0 +1,6 @@
+LDFLAGS=-w -extldflags
+
+build: ## Runs go build on repo
+	# Run go build and generate scorecard executable
+	CGO_ENABLED=0 go build -o scorecard-action -trimpath -a -tags netgo -ldflags '$(LDFLAGS)'
+
```
justaugustus, line 6:
```suggestion
	CGO_ENABLED=0 go build -o scorecard-action -trimpath -a -tags netgo -ldflags '$(LDFLAGS)'

```

I think the newline has an errant tab here.

entrypoint/entrypoint.go
```diff
@@ -60,11 +60,12 @@ func New() (*cobra.Command, error) {
...
```
justaugustus, line 64: non-blocking: by the time we get the results file, it should be complete and validated. That said, there's not much we can fix here (since the results file logic is in scorecard).

Leave a TODO here to take a look at this later.

options/env.go
```diff
@@ -31,14 +31,31 @@ const (
 	EnvGithubRepository        = "GITHUB_REPOSITORY"
 	EnvGithubRef               = "GITHUB_REF"
 	EnvGithubWorkspace         = "GITHUB_WORKSPACE"
-	EnvGithubAuthToken         = "GITHUB_AUTH_TOKEN" //nolint:gosec
+	EnvGithubAuthToken         = "INPUT_REPO_TOKEN" //nolint:gosec
```
justaugustus, line 34:
```suggestion
	EnvGithubAuthToken         = "GITHUB_AUTH_TOKEN" //nolint:gosec
	EnvInputRepoToken          = "INPUT_REPO_TOKEN" //nolint:gosec
```

Include both values.
If `GITHUB_AUTH_TOKEN` is present, then it is canonical.
We should only care about `INPUT_REPO_TOKEN` when `GITHUB_AUTH_TOKEN` is unset or empty.

Same comment for the `Env` struct below.

options/options.go
```diff
@@ -76,6 +76,11 @@ func New() (*Options, error) {
 	// Enable scorecard command to use SARIF format.
 	os.Setenv(options.EnvVarEnableSarif, trueStr)
 
+	// Parse relevant env vars.
+	if err := env.Parse(envVars); err != nil {
+		return nil, fmt.Errorf("parsing entrypoint env vars: %w", err)
+	}
```
justaugustus, line 82: Is there a reason to:
1. create a new struct (`Env`) which overlaps with `Options`
2. parse the env twice

See: https://github.com/ossf/scorecard-action/blob/e55a3ede2c4a2b684ec395a8f5ce054dcb929d61/options/options.go#L82-L84

Unless there's a good reason not to, any env vars that we need that are not already present should be added to `Options`.

options/options.go
```diff
@@ -91,6 +96,7 @@ func New() (*Options, error) {
 	}
 
 	// TODO(options): Move this set-or-default logic to its own function.
+	opts.ScorecardOpts.Format = envVars.EnvInputResultsFormat
```
justaugustus, line 99:
```suggestion
	opts.ScorecardOpts.Format = envVars.EnvInputResultsFormat
```

`INPUT_` vars should be last resort.
Default to SARIF and only override if `INPUT_RESULTS_FORMAT` is set AND populated.

options/options.go
```diff
@@ -107,6 +113,9 @@ func New() (*Options, error) {
 	}
 
 	opts.SetPublishResults()
+
+	opts.ScorecardOpts.ResultsFile = envVars.EnvInputResultsFile
```
justaugustus, line 117: Similar comment:
INPUT_ vars should be last resort.
Default to `SCORECARD_RESULTS_FORMAT` (`opts.ScoreOpts.Format`) and only prefer if it is unset or empty.

ref:
- https://github.com/ossf/scorecard/pull/1696
- https://github.com/ossf/scorecard/blob/c8acf3645f88b0dcbc5e798fe4b09730b869997b/options/options.go#L29-L52

justaugustus (changes requested):
@rohankh532 -- Left a few comments!

Dockerfile
```diff
@@ -21,23 +21,29 @@
...
+
 # TODO: use distroless.
+# FROM gcr.io/distroless/base:nonroot@sha256:02f667185ccf78dbaaf79376b6904aea6d832638e1314387c2c2932f217ac5cb
 FROM debian:11.2-slim@sha256:d5cd7e54530a8523168473a2dcc30215f2c863bfa71e09f77f58a085c419155b
```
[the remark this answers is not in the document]
justaugustus: Let's not make any changes to the images here.
I'd like to switch to ko + apko base, but will issue that in a separate PR once we do some more tidying.

ref: https://twitter.com/stephenaugustus/status/1499925583828078594

options/options.go
rohankh532: Originally this was to reduce the number of os calls in optoins.go, but adding them directly to the options struct seems like a better solution.

Dockerfile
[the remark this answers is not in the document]
rohankh532: Distroless was causing a permissions error when running os.Create in entrypoint.go to create the results file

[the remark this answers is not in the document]
rohankh532: Yeah, I reverted the Dockerfile change so it'll be addressed in another PR

justaugustus:
@rohankh532 -- I've got a few more review comments, but I'll issue them as pushes to your branch to speed things along!

justaugustus (approved):
Thanks @rohankh532!!

Issue #154, opened 2022-03-22
Using this action getting error 


marswed:
![image](https://user-images.githubusercontent.com/23168019/159508951-06b80f2e-a222-4ca3-a70d-e2784399dde1.png)

name: Scorecards supply-chain security
on:
  # Only the default branch is supported.
  branch_protection_rule:
  schedule:
     # Weekly on Saturdays.
    - cron: '30 1 * * 6'
    
  push:
    branches: [ master ]

# Declare default permissions as read only.
permissions: read-all

jobs:
  analysis:
    name: Scorecards analysis
    runs-on: ubuntu-latest
    permissions:
      # Needed to upload the results to code-scanning dashboard.
      security-events: write
      actions: read
      contents: read

    steps:
      - name: "Checkout code"
        #uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
        uses: actions/checkout@v2.4.0
        
        with:
          persist-credentials: false

      - name: "Run analysis"
        #uses: ossf/scorecard-action@c1aec4ac820532bab364f02a81873c555a0ba3a1 # v1.0.4
        uses: ossf/scorecard-action@v1.0.4
        
        with:
          results_file: results.sarif
          results_format: sarif
          # Read-only PAT token. To create it,
          # follow the steps in https://github.com/ossf/scorecard-action#pat-token-creation.
          #repo_token: ${{ secrets.SCORECARD_READ_TOKEN }}
          repo_token: ${{ secrets.SCORECARD2_READ_TOKEN || github.token }}
          
          # Publish the results to enable scorecard badges. For more details, see
          # https://github.com/ossf/scorecard-action#publishing-results.
          # For private repositories, `publish_results` will automatically be set to `false`,
          # regardless of the value entered here.
          publish_results: true

      # Upload the results as artifacts (optional).
      - name: "Upload artifact"
        #uses: actions/upload-artifact@82c141cc518b40d92cc801eee768e7aafc9c2fa2 # v2.3.1
        uses: actions/upload-artifact@v2.3.1
        with:
          name: SARIF file
          path: results.sarif
          retention-days: 5

      # Upload the results to GitHub's code scanning dashboard.
      - name: "Upload to code-scanning"
        #uses: github/codeql-action/upload-sarif@5f532563584d71fdef14ee64d17bafb34f751ce5 # v1.0.26
        uses: github/codeql-action/upload-sarif@v1.0.26
        with:
[...]

PR #155, opened 2022-03-22
✨ Add more info to logs

laurentsimon:
Add more info to logs to help troubleshooting problems.
There's been a new issue on a problem that we've seen before https://github.com/ossf/scorecard-action/issues/154

cc @naveensrinivasan let's add this to our list of unit tests once we've figured the problem.

```release-notes
Add more info to logs
```

closes https://github.com/ossf/scorecard-action/issues/156

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -47,7 +47,13 @@ export ENABLED_CHECKS=
 #
 # Boolean inputs are strings https://github.com/actions/runner/issues/1483.
 # ===============================================================================
-curl -s -H "Authorization: Bearer $GITHUB_AUTH_TOKEN" https://api.github.com/repos/$GITHUB_REPOSITORY > repo_info.json
+status_code=$(curl -s -H "Authorization: Bearer $GITHUB_AUTH_TOKEN" https://api.github.com/repos/"$GITHUB_REPOSITORY" -o repo_info.json -w '%{http_code}')
+if [[ $status_code -ge 400 ]]; then
+    error_msg=$(jq -r .message repo_info.json 2>/dev/null || echo 'unknown error')
+    echo "Failed to get repository information from GitHub, response $status_code: $error_msg"
+    exit 1;
+fi
+
 export SCORECARD_PRIVATE_REPOSITORY="$(cat repo_info.json | jq -r '.private')"
 export SCORECARD_DEFAULT_BRANCH="refs/heads/$(cat repo_info.json | jq -r '.default_branch')"
 export SCORECARD_IS_FORK="$(cat repo_info.json | jq -r '.fork')"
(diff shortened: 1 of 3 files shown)


naveensrinivasan (commented):
LGTM! Can we please also include this in the `go` code?

laurentsimon:
> LGTM! Can we please also include this in the `go` code?

IIUC the go code in out of sync already. Example; it's reading from the GITHUB_EVENT_PATH file https://github.com/ossf/scorecard-action/blob/main/options/options.go#L229 instead of using https://github.com/ossf/scorecard-action/blob/main/github/github.go#L115


I'll create an issue for sync'ing the code.

azeemshaikh38:
> LGTM! Can we please also include this in the `go` code?

+1 to this. It's easier to just do this as we go along. Even better, we can consider only adding updates to the Golang code instead of bash.

laurentsimon:
> > LGTM! Can we please also include this in the `go` code?
> 
> +1 to this. It's easier to just do this as we go along. Even better, we can consider only adding updates to the Golang code instead of bash.

I'll try that. I cannot add code only to go code though, because I have no idea if it does the same as entrypoint.sh. AT least it should be used by our own repos to stage a release.

laurentsimon:
please also advise how you'd like me to test. It used to be a file with the content, now it's an live API call to GitHub. I think I'll have to use the mock library; but tell me if you had something else in mind. Also, an example will be appreciated. @naveensrinivasan could you point me to a good example in scorecard you'd like me to follow?

azeemshaikh38:
> please also advise how you'd like me to test. It used to be a file with the content, now it's an live API call to GitHub. I think I'll have to use the mock library; but tell me if you had something else in mind. @naveensrinivasan could you point me to a good example in scorecard you'd like me to follow?

Do you mean test Golang code? @rohankh532 is working on adding e2e tests similar to how we test bash script today so can skip that for now. And the bash changes should get tested using the e2e we have already.

laurentsimon:
> > please also advise how you'd like me to test. It used to be a file with the content, now it's an live API call to GitHub. I think I'll have to use the mock library; but tell me if you had something else in mind. @naveensrinivasan could you point me to a good example in scorecard you'd like me to follow?
> 
> Do you mean test Golang code? @rohankh532 is working on adding e2e tests similar to how we test bash script today so can skip that for now. And the bash changes should get tested using the e2e we have already.

The code change in `go` now queries a GitHub API to retrieve repo information (instead of using a file content `GITHUB_EVENT_PATH`), so I need to mock the API if I want to have unit tests for it. The tests are failing right now, and I suspect it's because of that. I'm more than happy to leave this to @rohankh532... but that means `go` changes cannot be included in this PR... Wdut?

naveensrinivasan:
@laurentsimon  Is this still valid?

laurentsimon:
Let's close.

PR #158, opened 2022-03-30
Dockerfile Build Using Golang Entrypoint

rohankh532:
Changed the Dockerfile to use entrypoint.go instead of the shell script.

todo: add e2e tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,23 +21,30 @@
 #           -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN \
 #           -e GITHUB_REPOSITORY="ossf/scorecard" \
 #           laurentsimon/scorecard-action:latest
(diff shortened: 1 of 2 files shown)


azeemshaikh38 (approved):

azeemshaikh38:
@naveensrinivasan could you help with setting up e2e test for this? Basically, just like we have e2e tests today pointing to `@main`, can have a separate workflow in the test repos pointing to `@golang-staging` branch. That way, it'll be easy to figure out any breakages or differences between the 2 versions.

azeemshaikh38:
> > @naveensrinivasan could you help with setting up e2e test for this? Basically, just like we have e2e tests today pointing to `@main`, can have a separate workflow in the test repos pointing to `@golang-staging` branch. That way, it'll be easy to figure out any breakages or differences between the 2 versions.
> 
> Sounds good. I can do that early next week.

Thanks Naveen!

PR #170, opened 2022-04-09
📖 docs/e2e: Add information about golang-staging branch tests

naveensrinivasan:
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>


azeemshaikh38 (dismissed):
This is only needed while we rollout the Golang action right? Can have this info on the issue which is being used to track the rollout instead.

justaugustus (approved):
> This is only needed while we rollout the Golang action right? Can have this info on the issue which is being used to track the rollout instead.

Merging this because there is currently no easily discoverable information in the repo about the Golang staging testing.
Not a big deal to remove once the branch is no longer required.

PR #171, opened 2022-04-11
Default Branch Checking Bugfix

### LINKED ISSUE #169: Failed to run automated scorecard-action@main ossf-tests/scorecard-action-non-main-branch
https://github.com/ossf-tests/scorecard-action-non-main-branch/actions/runs/2142139482

rohankh532:
PR to fix #169


naveensrinivasan (approved):
Looks good. @azeemshaikh38 👀

naveensrinivasan:
Shouldn't this target golang-staging branch?

rohankh532:
> Shouldn't this target golang-staging branch?

Oops, fixed.

azeemshaikh38:
> > Shouldn't this target golang-staging branch?
> 
> Oops, fixed.

This should not be `golang-staging`, merging this to `main` is the right thing to do. Copying my reply to Naveen over slack here: 

> we want to merge these changes to main  and after that to golang-staging . Basically, the only diff between main  and golang-staging  should be that the Dockerfile uses Golang instead of Bash. Everything else should be equivalent

rohankh532:
> > > Shouldn't this target golang-staging branch?
> > 
> > 
> > Oops, fixed.
> 
> This should not be `golang-staging`, merging this to `main` is the right thing to do. Copying my reply to Naveen over slack here:
> 
> > we want to merge these changes to main  and after that to golang-staging . Basically, the only diff between main  and golang-staging  should be that the Dockerfile uses Golang instead of Bash. Everything else should be equivalent

Got it, changed back to main.

Dockerfile
```diff
@@ -21,23 +21,30 @@
...
+WORKDIR /src
+ENV CGO_ENABLED=0
+COPY go.* ./
+RUN go mod download
+COPY . ./
+
+FROM base AS build
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 make build
```
azeemshaikh38, line 36: Dockerfile shouldn't be updated?

azeemshaikh38 (changes requested):
LGTM overall. Revert the changes on Dockerfile and ensure that the unit tests are passing.

Dockerfile
rohankh532: Reverted this

azeemshaikh38 (approved):
Thanks!

Issue #180, opened 2022-04-15
End-to-end tests for GITHUB_TOKEN


laurentsimon:
Add e2e tests using the default GITHUB_TOKEN

/cc @naveensrinivasan

naveensrinivasan:
Here are actions with GitHub Token



Public repo - https://github.com/ossf-tests/scorecard-action/blob/6acbb7d090d1cfff7c51c58b8fc501c87e327995/.github/workflows/scorecards.yml#L35

Private repo - https://github.com/test-organization-ls/scorecard-action-private-repo-tests/blob/577c5bb3ceb1c636b277ae159a0ef115c1c0c9fd/.github/workflows/scorecard.yml#L35

laurentsimon:
Thanks. I'm going to make a change to the action so that the GITHUB_TOKEN is the default value if not token is provided by the user. I'll cc you on the PR. Once it's merged, can we add 2 more tests with no token provided, to verify it all works as well?

naveensrinivasan:
> Thanks. I'm going to make a change to the action so that the GITHUB_TOKEN is the default value if not token is provided by the user. I'll cc you on the PR. Once it's merged, can we add 2 more tests with no token provided, to verify it all works as well?

Will do that!

naveensrinivasan:
### Main branch without any token
https://github.com/ossf-tests/scorecard-action/actions/runs/2192902943
https://github.com/ossf-tests/scorecard-action/blob/d955b923ca38d81e755a8d77533f121cf5a8dbb3/.github/workflows/scorecards-without-token.yml#L36

### Private repo without any token
https://github.com/test-organization-ls/scorecard-action-private-repo-tests/actions/runs/2192919167
https://github.com/test-organization-ls/scorecard-action-private-repo-tests/blob/76cee4494a9e2b62b22417386a4a219b47e56bec/.github/workflows/scorecards-without-token.yml#L35

laurentsimon:
All working, great!
Are all these new tests documented in https://github.com/ossf/scorecard-action/tree/main/e2e?

I think they are, just double-checking. I wonder if README in the 2 ossf-tests repos should also list what each workflow does. Nothing fancy, just something to speed up maintainers when adding tests.

Thanks again for all this @naveensrinivasan getting closer to our ext release!

laurentsimon:
Is https://github.com/ossf/scorecard-action/tree/main/e2e failing? One of the tests is red

laurentsimon:
[...]

PR #185, opened 2022-04-18
✨ Set default token as GITHUB_TOKEN

laurentsimon:
Set default token to be GITHUB_TOKEN.
Users don't need to provide it, unless they want to get the Branch-Protection results.

I'll update the README in a separate PR.

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -30,7 +30,7 @@ inputs:
   repo_token:
     description: "INPUT: GitHub token with read access"
     required: false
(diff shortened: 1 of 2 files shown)


laurentsimon:
@naveensrinivasan after merging this PR, we can address https://github.com/ossf/scorecard-action/issues/180#issuecomment-1101544006

naveensrinivasan (approved):

PR #194, opened 2022-04-21
Add initial release documentation

justaugustus:
Captures the content from https://github.com/ossf/scorecard-action/issues/33.
~I'll clean this up a little before removing the `[WIP]` tag.~


RELEASE.md
```diff
@@ -0,0 +1,22 @@
...
+[tracking issue](https://github.com/ossf/scorecard-action/issues/33) or issue a
+pull request to discuss.)
+
+- [steps:](#steps)
+
+## steps:
+
+1. cut a scorecard release and wait for a container image to be created and tagged with new release. Note the hash of the container as `CH1`. Note: we do not need a scorecard release, we can use any stable version we want.
+2. update the hash pin in our [dockerfile](https://github.com/ossf/scorecard-action/blob/main/.github/Dockerfile#L26) to use the container hash `CH1` from step 1. Once the PR is merged, note the GitHub's commit hash as `GH2`.
+~~3. [manually trigger](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow) the workflow to generate our container image. Note the hash of the container image generated as `CH3`. It can be found [here](https://github.com/ossf/scorecard-action/pkgs/container/scorecard-action) using the manifest's "digest"`.~~
```
laurentsimon, line 16: step 3 and 4 are remnants of the past, they are no longer needed.

RELEASE.md
```diff
@@ -0,0 +1,22 @@
...
+
+(If there are improvements you'd like to see, please comment on the
+[tracking issue](https://github.com/ossf/scorecard-action/issues/33) or issue a
+pull request to discuss.)
+
+- [steps:](#steps)
+
+## steps:
+
+1. cut a scorecard release and wait for a container image to be created and tagged with new release. Note the hash of the container as `CH1`. Note: we do not need a scorecard release, we can use any stable version we want.
```
laurentsimon, line 14: none of these hashes are used in scripts.

CH1 is the Container image's Hash for the scorecard container (scorecard repo). There's one container generated for every stable build (after it's gone thru integration tests) and for each release. We use this container as part of a multi-stage build to pull the scorecard binary in https://github.com/ossf/scorecard-action/blob/main/Dockerfile#L34

HG2 means Github Hash: it's the commit hash of this repo (scorecard-action) after we've updated the docker image to use the scorecard container image we want.

justaugustus: Thanks Laurent.
I just wanted to make sure!

RELEASE.md
justaugustus: Removed!

RELEASE.md
```diff
@@ -0,0 +1,115 @@
...
+be merged before releasing the scorecard GitHub Action.
+
+### Update the scorecard version
+
+_NOTE: As the scorecard GitHub Action is based on scorecard, you may want to publish a new release of scorecard to ensure the next release of the GitHub Action has the most up-to-date functionality. This is not strictly required. The only requirement is that we use a stable scorecard version which is at or above the current version used for this action._
+
+For the rest of document, let `CH1` be the hash of the scorecard image you
+intend to use for this release.
+
+See [here](https://github.com/orgs/ossf/packages?repo_name=scorecard) for
```
laurentsimon, line 45: I don't think we release on GH registry. I think it's gcr.io/openssf/scorecard instead. Should be accessible by anyone

laurentsimon (approved):
LGTM.

RELEASE.md
justaugustus: Let's address in a follow-up.
ref: https://github.com/ossf/scorecard-action/issues/33#issuecomment-1108574853

Issue #196, opened 2022-04-22
Add full log of the repo_json file


laurentsimon:
We need to log the response from https://github.com/ossf/scorecard-action/blob/main/entrypoint.sh#L50, because there seems to be intermittent errors when querying this endpoint which are hard to fix: see https://github.com/ossf/scorecard-action/issues/154

@rohankh532 are you interested in taking this one?

/cc @azeemshaikh38

laurentsimon:
@rohankh532 are you interested in taking it as your farewell :-)?

laurentsimon:
closed by https://github.com/ossf/scorecard-action/pull/211

PR #211, opened 2022-04-27
Log repo_info.json File in entrypoint.sh

rohankh532:
For issue #196


entrypoint.sh
```diff
@@ -47,7 +47,13 @@ export ENABLED_CHECKS=
 #
 # Boolean inputs are strings https://github.com/actions/runner/issues/1483.
 # ===============================================================================
-curl -s -H "Authorization: Bearer $GITHUB_AUTH_TOKEN" https://api.github.com/repos/$GITHUB_REPOSITORY > repo_info.json
+status_code=$(curl -s -H "Authorization: Bearer $GITHUB_AUTH_TOKEN" https://api.github.com/repos/"$GITHUB_REPOSITORY" -o repo_info.json -w '%{http_code}')
+if [[ $status_code -ge 300 ]]; then
```
laurentsimon, line 51: also check for <200

entrypoint.sh
```diff
@@ -47,7 +47,13 @@ export ENABLED_CHECKS=
...
+    error_msg=$(jq -r .message repo_info.json 2>/dev/null || echo 'unknown error')
+    echo "Failed to get repository information from GitHub, response $status_code: $error_msg"
+    exit 1;
+fi
+
 export SCORECARD_PRIVATE_REPOSITORY="$(cat repo_info.json | jq -r '.private')"
 export SCORECARD_DEFAULT_BRANCH="refs/heads/$(cat repo_info.json | jq -r '.default_branch')"
 export SCORECARD_IS_FORK="$(cat repo_info.json | jq -r '.fork')"
```
laurentsimon, line 59: we want to print the entire repo_info.json to help troubleshooting, then we can remove the file

entrypoint.sh
laurentsimon: example https://github.com/slsa-framework/slsa-github-generator/pull/38/files

entrypoint.sh
rohankh532: fixed

entrypoint.sh
```diff
@@ -47,7 +47,14 @@ export ENABLED_CHECKS=
 #
 # Boolean inputs are strings https://github.com/actions/runner/issues/1483.
 # ===============================================================================
-curl -s -H "Authorization: Bearer $GITHUB_AUTH_TOKEN" https://api.github.com/repos/$GITHUB_REPOSITORY > repo_info.json
+status_code=$(curl -s -H "Authorization: Bearer $GITHUB_AUTH_TOKEN" https://api.github.com/repos/"$GITHUB_REPOSITORY" -o repo_info.json -w '%{http_code}')
+if [[ $status_code -lt 200 ]] || [[ $status_code -ge 300 ]]; then
+    error_msg=$(jq -r .message repo_info.json 2>/dev/null || echo 'unknown error')
+    echo "Failed to get repository information from GitHub, response $status_code: $error_msg"
+    echo "$(<repo_info.json)"
```
laurentsimon, line 54: this looks clever :-) `cat "$file"` would have worked too I think, Want to remove the file?

laurentsimon (approved):

entrypoint.sh
laurentsimon, line 54: note: probably fine to display the content all the times, not just on failure. I think what you have works, but I'm always worried about some unexpected corner cases where the data may still be valuable

laurentsimon (approved):

laurentsimon (approved):
Thanks

PR #247, opened 2022-05-06
✨ Update Scorecard version to v4.2.0 in Golang

azeemshaikh38:
Update the Golang Scorecard version to the latest - v4.2.0


naveensrinivasan (approved):
Thank you!

Issue #277, opened 2022-05-10
BUG: gostaging


laurentsimon:
gostaging is broken due to https://github.com/ossf/scorecard/pull/1898.

One way to fix it is for the go code to grab env variables and call scorecard CLI/API with the correct options set. This decouples scorecard CLI and the Action code.

justaugustus:
(Discussed [here](https://openssf.slack.com/archives/C0235AR8N2C/p1653431862505869?thread_ts=1653419973.820599&cid=C0235AR8N2C): I'll take care of fixing this break)

justaugustus:
Should be fixed with the merge of https://github.com/ossf/scorecard-action/pull/313, but feel free to reopen, if needed!

Issue #236, opened 2022-05-03
SARIF difference in golang-staging


laurentsimon:
See results https://github.com/ossf-tests/scorecard-action-results/pull/2
Workflow that generated it is https://github.com/ossf-tests/scorecard-action/runs/6187072538?check_suite_focus=true

The results are odd and wrong. Consider https://github.com/ossf-tests/scorecard-action-results/blob/a61cd2ab16265c0dad77b7fc9eb3fa0e7f7fa6b3/scorecard-action-main.sarif#L271-L311 or more specifically https://github.com/ossf-tests/scorecard-action-results/blob/a61cd2ab16265c0dad77b7fc9eb3fa0e7f7fa6b3/scorecard-action-main.sarif#L274

We see that the text contains both the scorecard detail text and the location information, concatenated: `score is 8: dependency not pinned by hash detected -- score normalized to 8:\nWarn: third-party action not pinned by hash: .github/workflows/scorecards-golang.yml:`

I ran scorecard at HEAD and at the v4.1.0 release on the same repo (`ossf-tests/scorecard-action`), and the results are identical but are **different**  from the results of the run above for golang-staging:
```
{
          "ruleId": "PinnedDependenciesID",
          "ruleIndex": 4,
          "message": {
            "text": "third-party action not pinned by hash\nClick Remediation section below to solve this issue"
          },
          "locations": [
            {
              "physicalLocation": {
                "region": {
                  "startLine": 30,
                  "endLine": 30,
                  "snippet": {
                    "text": "ossf/scorecard-action@golang-staging"
                  }
                },
                "artifactLocation": {
                  "uri": ".github/workflows/Scorecards-with-default-GH-Token-golang-staging.yml",
                  "uriBaseId": "%SRCROOT%"
                }
              },
              "message": {
                "text": "third-party action not pinned by hash"
              }
            }
          ]
        },
```

You see that the text message only contains the text and the file-based results appear within the `locations` instead.

@naveensrinivasan can you confirm I'm understanding this correctly? It's possible I missed something

PR #316, opened 2022-05-25
image: Build the `scorecard` command from this repo's wrapper instead of copying

justaugustus:
PRs like https://github.com/ossf/scorecard-action/pull/312 and https://github.com/ossf/scorecard-action/pull/313 demonstrate that it's possible to drift between what is built for the scorecard action container image and what dependencies are in the Golang-based action, which means the behavior can be unpredictable and can lead to breakages, as seen in https://github.com/ossf/scorecard-action/issues/277.

This PR is a small change, which ensures we only need to update the scorecard dependency in a single place.
I also added some baseline testing docs, as a treat!

Commits:

- image: Build the `scorecard` command from source instead of copying

  This is another step in getting rid of the bash-based entrypoint.
  By building from the scorecard-action source, we're ensuring a few
  things:

  - The existing entrypoint leverages the wrapper code
  - We only track the scorecard dependency in a single location (go.mod)

- image: Include event paths for testing scenarios

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 # Testing
 unit-coverage.out
 scorecard-action
+output/
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,16 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Testing: docker run -e GITHUB_REF=refs/heads/main \
-#           -e GITHUB_EVENT_NAME=branch_protection_rule \
-#           -e INPUT_RESULTS_FORMAT=sarif \
-#           -e INPUT_RESULTS_FILE=results.sarif \
-#           -e GITHUB_WORKSPACE=/ \
-#           -e INPUT_POLICY_FILE="/policy.yml" \
-#           -e INPUT_REPO_TOKEN=$GITHUB_AUTH_TOKEN \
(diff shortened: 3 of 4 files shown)


justaugustus:
(I'm going to save any thoughts I had about being clever with the bash script for a future PR to keep the diff small.)

laurentsimon:
One worry is time to run the Action. Compiling scorecard takes several minutes, so it would be great to use pre-build containers. We can compile the container using the Dockefile you've updated.  

How about temporarily renaming this Dockefile to Dockerfile.go? Otherwise the e2e tests will start breaking immediately. We have this issue https://github.com/ossf/scorecard-action/issues/236 that needs to be resolved before we can move to the go code

Wdut?

laurentsimon:
One worry is time to run the Action. Compiling scorecard takes several minutes, so it would be great to use pre-build containers. We can compile the container using the Dockefile you've updated.  

How about temporarily renaming this Dockefile to Dockerfile.go? Otherwise the e2e tests will start breaking immediately. We have this issue https://github.com/ossf/scorecard-action/issues/236 that needs to be resolved before we can move to the go code

Wdut?

azeemshaikh38:
The Dockerfile changes should be pushed to `golang-staging` branch. That'll automatically test this change for Golang action against the current bash action. Apart from that LGTM.

spencerschrock:
This PR is old and appears to have been addressed by https://github.com/ossf/scorecard-action/pull/750

justaugustus:
> This PR is old and appears to have been addressed by https://github.com/ossf/scorecard-action/pull/750

Thanks for catching this and cleaning up, @spencerschrock!

PR #337, opened 2022-06-06
📖 docs: Add instructions on building/testing container image

justaugustus:
(Pulled these file changes from https://github.com/ossf/scorecard-action/pull/316, since that PR is still pending)

- Incorporates container image build/test docs into repo documentation
- Update references in `Dockerfile` to point to the build/test docs
- Adds a .dockerignore file (which is a symlink to .gitignore)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+.gitignore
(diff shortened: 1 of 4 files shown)


azeemshaikh38 (approved):
Thanks!

laurentsimon (approved):

PR #414, opened 2022-06-08
🐛 Avoid error on `json` and `default` inputs

### LINKED ISSUE #386: BUG: Bash action fails on `json` and `default` input formats
See #382, #383, #384 and #385

azeemshaikh38:
Referring to an unset var causes `unbound variable` error. Check if the var is set before we refer to it. Fixes #386.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -78,7 +78,9 @@ echo "Fork repository: $SCORECARD_IS_FORK"
 echo "Private repository: $SCORECARD_PRIVATE_REPOSITORY"
 echo "Publication enabled: $SCORECARD_PUBLISH_RESULTS"
 echo "Format: $SCORECARD_RESULTS_FORMAT"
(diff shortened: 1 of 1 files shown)


azeemshaikh38:
> When is the policy file not set? In the Action, it has to always be there to generate the SARIF results.
> 
> Is this some corner case for testing?

It's unset here: https://github.com/ossf/scorecard-action/blob/0b3cf46f654c5617af003e05d307898eeb0b6481/entrypoint.sh#L69-L71 when input results_format is not sarif.

laurentsimon:
> > When is the policy file not set? In the Action, it has to always be there to generate the SARIF results.
> > Is this some corner case for testing?
> 
> It's unset here:
> 
> https://github.com/ossf/scorecard-action/blob/0b3cf46f654c5617af003e05d307898eeb0b6481/entrypoint.sh#L69-L71
> 
> when input results_format is not sarif.

gotcha. Nobody uses this yet, but good catch

PR #766, opened 2022-07-15
🌱 Prepare for v2 beta1 release

azeemshaikh38:
Commit created from this PR will be used to generate the pre-release of scorecard-action. I've decided to name the tag - `v2.0.0-beta.1`.

PR #771, opened 2022-07-20
🐛 Fix dependency conflicts in go.mod

azeemshaikh38:
Fixes the dependency conflicts we were seeing when running `go mod tidy`. Took my cue from [this `cosign` PR](https://github.com/sigstore/cosign/pull/2070) to find the issue.  Should also fix the dependabot security alert and unblock v2 beta release - https://github.com/ossf/scorecard-action/pull/766

There were some linter failures that I needed to fix too in this PR. Our CI is very broken currently and we need a single PR to fix all of them to get the PR merges to happen.


naveensrinivasan (approved):
Thanks!

PR #773, opened 2022-07-25
Organization-wide Workflow Add Instructions

naveensrinivasan:
I picked up the PR #151

PR #776, opened 2022-07-26
multi-repo-action: Note that tool is a work-in-progress

naveensrinivasan:
Based on the discussion from this https://github.com/ossf/scorecard-action/pull/773


justaugustus (approved):
Thanks @naveensrinivasan!

PR #779, opened 2022-07-27
📖 Update README about the restrictions for scorecard-action:v2

azeemshaikh38:
scorecard-action:v2 only works with certain approved GH actions. Clearly document this restriction for users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,15 @@ The Scorecards GitHub Action is free for all public repositories. Private reposi
 
 ## Breaking changes in v2
 
-Starting from scorecard-action:v2, `GITHUB_TOKEN` permissions needs to incude
-`token_id: write` for `publish_results: true`. This is needed to access GitHub's
-OIDC token whuch verifies the authenticity of the result when publishing it.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,10 +8,15 @@ The Scorecards GitHub Action is free for all public repositories. Private reposi
...
 
-Starting from scorecard-action:v2, `GITHUB_TOKEN` permissions needs to incude
-`token_id: write` for `publish_results: true`. This is needed to access GitHub's
-OIDC token whuch verifies the authenticity of the result when publishing it.
-
+Starting from scorecard-action:v2, `GITHUB_TOKEN` permissions or job permissions needs to incude
+`id-token: write` for `publish_results: true`. This is needed to access GitHub's
+OIDC token which verifies the authenticity of the result when publishing it.
+
+scorecard-action:v2 also requires that the steps in the job running ossf/scorecard-action step
```
laurentsimon, line 15: hopefully that is the general case, since the example workflows do just that.

laurentsimon (approved):

README.md
azeemshaikh38: Yeah hope so too. Would be good to make this generic enough and fix it in the right way though in the long run.

PR #791, opened 2022-08-14
🐛 Add request application/json request header

azeemshaikh38:
Explicitly adds `application/json` request header. API calls with stricter verification fail without this header.


naveensrinivasan (approved):
Thanks!

Issue #801, opened 2022-08-15
BUG: Failing API calls to the OpenAPI/Swagger based endpoint


azeemshaikh38:
Related: #792, #793, #794, #795, #796, #798, #799, #800

OpenAPI/Swagger based APIs expect the request to explicitly specify `application/json` as request header. To fix this I plan to:

- Rollback the current change so existing users of beta scorecard-action are not affected.
- Update the e2e tests to talk to the staging API instead of prod, that way we catch these issues before it goes into production.
- Update scorecard-action calls to use `application/json`.

azeemshaikh38:
This is now fixed. Keeping it open till I get the e2e tests working.

azeemshaikh38:
@ossf/scorecard-maintainers need some input - for the e2e tests to work on staging API url (so that we catch bugs before they make it into production), I need a way to override the publish URL. I added an env override only to realize that the workflow verification fails (we don't allow env vars to be set in scorecard workflows for result integrity). So thinking of adding `publish_base_url` to `action.yaml`. What do folks think?

laurentsimon:
SGTM.

laurentsimon:
Let's just hide this variable behind a `internal-publish-base-url`, so that there's no expectation from users.

PR #803, opened 2022-08-15
Create a new release v2.0.0-alpha.1

azeemshaikh38:
The fixes made in #791 need to be released to users. So creating a alpha release.

This PR also allows for the API url to be overridden. Used in e2e tests.

Part of #801.


naveensrinivasan (approved):
Thanks

PR #845, opened 2022-08-30
Fix workflow path in automatic creation of PR 

RadoslavGatev:
The file is wrongly created at `/scorecards.yml` instead of `.github/workflows/scorecards.yml`.

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -43,8 +43,9 @@ To report any issues with this tool, see [here](https://github.com/ossf/scorecar
 `
 
 	pullRequestTitle = commitMessage
+	workflowFilePath = path.Join(workflowBase, workflowFile)
 	workflowFiles    = []string{
-		path.Join(workflowBase, workflowFile),
+		workflowFilePath,
 		path.Join(workflowBase, workflowFileDeprecated),
 	}
 )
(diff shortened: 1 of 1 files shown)


naveensrinivasan (commented):
Could you please add some tests?

azeemshaikh38:
@RadoslavGatev could you clarify what this PR fixes? The installer tool is supposed to commit the file to `.github/workflows/` directory since this is a GitHub workflow.

RadoslavGatev:
@naveensrinivasan I don't see a test suite for the installer. If it is there I might add a test case but I don't have the availability to implement it from scratch. 

@azeemshaikh38 It is indeed supposed to commit the file to `.github/workflows/` but it used to commit it to the root of the repository (as the path to the `workflows` folder was not prepended). This is what this PR is for.

azeemshaikh38:
> @naveensrinivasan I don't see a test suite for the installer. If it is there I might add a test case but I don't have the availability to implement it from scratch.
> 
> @azeemshaikh38 It is indeed supposed to commit the file to `.github/workflows/` but it used to commit it to the root of the repository (as the path to the `workflows` folder was not prepended). This is what this PR is for.
> 
> PS I just saw that the initial PR description is a bit misleading. :)

Got it, that makes sense. Thanks!

azeemshaikh38 (approved):

azeemshaikh38:
@justaugustus fyi

justaugustus (approved):
Thanks for catching this, @RadoslavGatev!

> Could you please add some tests?

@naveensrinivasan -- I'll take care of this before release.

PR #846, opened 2022-08-30
[WIP] Install subcommand

justaugustus:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -25,7 +25,6 @@ linters:
   enable:
     - asciicheck
     - bodyclose
-    - deadcode
     - depguard
     - dogsled
     - errcheck
@@ -64,7 +63,6 @@ linters:
     - rowserrcheck
     - sqlclosecheck
     - staticcheck
-    - structcheck
     - stylecheck
     - thelper
     - tparallel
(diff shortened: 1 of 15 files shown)


azeemshaikh38:
@justaugustus is this ready for review?

azeemshaikh38 (changes requested):
IMO, what is really needed to unblock us from launching this tool is its correctness/readiness. Some things which need fixing:

- When I tried running this on my repo (https://github.com/azeemshaikh38/CacheSimulator), the tool just failed.
- There are no docs mentioning that `GITHUB_AUTH_TOKEN` env var needs to be set. 
- There should be sample PRs to show how the tool behavior will look like.
- We are simply checking for existence for `scorecards.yml` and `scorecard-analysis.yml`. This is fine for now, but can be buggy. So either (i) document that behavior for users, or (ii) remove it completely and we tell users we dumbly create PR every time.
- The branch name that is created could exist before and we risk overwriting someone else's branch. Probably can be something more unique like `scorecard-<current_time>`.

Apart from the issues above, I'm also hesitant to sign off on this PR since this PR touches the action code and we're a few days away from launch. There is also a long weekend coming up so some of us maybe inaccessible. 

Why not keep this separate from the action (in its own Go module) for now and we release it as a separate tool from the action?

naveensrinivasan:
>Apart from the issues above, I'm also hesitant to sign off on this PR since this PR touches the action code and we're a few days away from launch. There is also a long weekend coming up so some of us maybe inaccessible.

+1

PR #849, opened 2022-09-01
📖 Add docs for API

azeemshaikh38:
PR adds docs to show REST API usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ ________
 - [Workflow Setup](#workflow-setup)
 
 [View Results](#view-results)
+- [REST API](#rest-api)
 - [Scorecard Badge](#scorecard-badge)
 - [Code Scanning Alerts](#code-scanning-alerts)
 - [Verify Runs](#verify-runs)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -110,6 +110,9 @@ Then click "Add More Scanning Tools."
 
 The workflow is preconfigured to run on every repository contribution. After making a code change, you can view the results for the change either through the Scorecard Badge, Code Scanning Alerts or GitHub Workflow Runs.
 
+### REST API
+Starting with scorecard-action:v2, users can use a REST API to query their latest run results. This requires setting `publish_results: true` for the action and enabling `id-token: write` permission for the job (needed to access GitHub OIDC token). The API is available here: https://api.securityscorecards.dev.
```
naveensrinivasan, line 114: How about including a link to an example like this for `publish_results` https://github.com/ossf/scorecard/blob/d13ba3f3355b958d5d62edc47282a2e7ed9fa7c1/.github/workflows/scorecard-analysis.yml#L39 and for `id-token` https://github.com/ossf/scorecard/blob/d13ba3f3355b958d5d62edc47282a2e7ed9fa7c1/.github/workflows/scorecard-analysis.yml#L22?

This makes it clear for anyone new.

naveensrinivasan (approved):

README.md
azeemshaikh38: Done

laurentsimon (approved):
Nice! So I think we're ready to move forward with the dependency visualization on pull requests?

azeemshaikh38:
> Nice! So I think we're ready to move forward with the dependency visualization on pull requests?

I think so. Maybe after the launch is completed, but yes.

Issue #856, opened 2022-09-08
githubv4.Query: Resource not accessible by integration


fmarier:
I'm using the [default GitHub Action workflow](https://github.com/ossf/scorecard-action#workflow-example), though with tags instead of commit hashes in https://github.com/brave/simplepadding/blob/master/.github/workflows/scorecards.yml and I'm seeing this [error](https://github.com/brave/simplepadding/runs/8257646939?check_suite_focus=true) now that I've upgraded to v2:
```
Error: one or more checks had a runtime error: internal error: error during branchesHandler.setup: internal error: githubv4.Query: Resource not accessible by integration
```
Is there a permission missing from the example workflow?

I see it's also been reported in https://github.com/ossf/scorecard/issues/1097, but I'm not sure I'm seeing the same issue since it was working fine before I [upgraded from v1.1.2 to v2](https://github.com/brave/simplepadding/commit/d6632a8a3487d10c4e1b8b4bafdaf609c31c4771).

azeemshaikh38:
Thanks for the report. I can reproduce this on my end. Seems like switching on specific branch-protection settings (like `Enable signed commits`) can make the action throw up this error. I haven't tested exhaustively to see what all settings lead to this error, but for now I'll try to send out a fix to ensure Scorecard exists gracefully in such cases.

fmarier:
Here are the branch protection settings I have enabled in the above repo:
![Screenshot from 2022-09-08 17-24-37](https://user-images.githubusercontent.com/167821/189247914-9eb8c51c-b7b2-4c0d-b017-7cb0857fab7d.png)

nijel:
It happens for https://github.com/WeblateOrg/weblate/ as well, the only branch protection we have enabled is "Require status checks to pass before merging". Failing workflow run: https://github.com/WeblateOrg/weblate/runs/8262284657

dirien:
Not adding additional noise, but happens in https://github.com/dirien/minectl-sdk too.

https://github.com/dirien/minectl-sdk/runs/8267086312?check_suite_focus=true

Only change is the update to 2.0.0.

dirien:
I compared a the output from cosign (Ping @cpanato) and my and see following difference by same  pipline file:


cosign: 
```bash
Repository: sigstore/cosign
Fork repository: false
Private repository: false
Publication enabled: false
Format: json
Policy file: 
Default branch: main
Using payload from: results.json
```

the failed one:
```bash
Private repository: false
Publication enabled: false
[...]

PR #948, opened 2022-09-26
🐛 Restore behavior of ignoring scorecard runtime errors

### LINKED ISSUE #269: E2e tests improvements
- [ ] Re-use existing issue instead of creating new one (Disable gostaging until this is done: we're inundated with issues otherwise)
- [x] Verify push/dispatch_workflow work properly
- [ ] Auto-close issues when e2e tests pass
- [x] Fix the content which contains `\n` instead of displaying a new line
- [x] Links to workflow files are broken
- [x] Provide the token type in the issue description (PAT or GITHUB_TOKEN)

spencerschrock:
This is a proper fix for #856.

As of [scorecard v4.7.0](https://github.com/ossf/scorecard/releases/tag/v4.7.0), runtime errors encountered during checks now produce a `ErrorCheckRuntime` error. These errors could be expected, for example repos with branch protection enabled but did not provide a PAT (currently a known limitation of scorecard). Or they could be from unexpected conditions / secondary rate limits.

These errors started bubbling up in scorecard v4.6.0, and while this was beneficial for CLI users of scorecard it broke the existing behavior of scorecard-action. To restore the behavior, scorecard-action now checks for these specific scorecard errors and continues to silently ignore them. 

e2e tests which would have caught #856 have been added in github.com/ossf-tests/scorecard-action-branch-protection-e2e
This also resolves #269

### DIFF
--- a/entrypoint/entrypoint.go
+++ b/entrypoint/entrypoint.go
@@ -83,11 +83,10 @@ func New() (*cobra.Command, error) {
 	// users were having action failures due to secondary rate limits causing checks to fail
 	scorecardRunE := actionCmd.RunE
 	actionCmd.RunE = func(cmd *cobra.Command, args []string) error {
-		err := scorecardRunE(cmd, args)
-		if errors.Is(err, sce.ErrorCheckRuntime) {
-			err = nil
+		if err := scorecardRunE(cmd, args); err != nil && !errors.Is(err, sce.ErrorCheckRuntime) {
+			return err
 		}
-		return err
+		return nil
 	}
 
 	actionCmd.PersistentPostRun = func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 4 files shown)


entrypoint/entrypoint.go
```diff
@@ -78,6 +79,17 @@ func New() (*cobra.Command, error) {
...
 
+	// wrap scorecard's existing RunE, but ignore runtime errors that occur during checks
+	// users were having action failures due to secondary rate limits causing checks to fail
+	scorecardRunE := actionCmd.RunE
+	actionCmd.RunE = func(cmd *cobra.Command, args []string) error {
+		err := scorecardRunE(cmd, args)
+		if errors.Is(err, sce.ErrorCheckRuntime) {
+			err = nil
+		}
+		return err
```
azeemshaikh38, line 90: Nit:

```
if err := scorecardRunE(cmd, args); err != nil && !errors.Is(err, sce.ErrorcheckRuntime) {
  return err
}
return nil
```

azeemshaikh38 (approved):
Thanks!

Issue #978, opened 2022-10-18
Docker tag scorecard-action:v2.0.5 not found 


konstruktoid:
Renovate PR: "Update ossf/scorecard-action action to v2.0.5" (https://github.com/konstruktoid/ansible-role-hardening/pull/185)

https://github.com/konstruktoid/ansible-role-hardening/actions/runs/3276257156/jobs/5393791624#step:2:1

```console
Pull down action image 'gcr.io/openssf/scorecard-action:v2.0.5'
  /usr/bin/docker pull gcr.io/openssf/scorecard-action:v[2](https://github.com/konstruktoid/ansible-role-hardening/actions/runs/3276257156/jobs/5393791624#step:2:2).0.5
  Error response from daemon: manifest for gcr.io/openssf/scorecard-action:v2.0.5 not found: manifest unknown: Failed to fetch "v2.0.5" from request "/v2/openssf/scorecard-action/manifests/v2.0.5".
  Warning: Docker pull failed with exit code 1, back off 5.34 seconds before retry.
  /usr/bin/docker pull gcr.io/openssf/scorecard-action:v2.0.5
  Error response from daemon: manifest for gcr.io/openssf/scorecard-action:v2.0.5 not found: manifest unknown: Failed to fetch "v2.0.5" from request "/v2/openssf/scorecard-action/manifests/v2.0.5".
  Warning: Docker pull failed with exit code 1, back off [3](https://github.com/konstruktoid/ansible-role-hardening/actions/runs/3276257156/jobs/5393791624#step:2:3).379 seconds before retry.
  /usr/bin/docker pull gcr.io/openssf/scorecard-action:v2.0.5
  Error response from daemon: manifest for gcr.io/openssf/scorecard-action:v2.0.5 not found: manifest unknown: Failed to fetch "v2.0.5" from request "/v2/openssf/scorecard-action/manifests/v2.0.5".
Error: Docker pull failed with exit code 1
```

naveensrinivasan:
This PR broker this https://github.com/ossf/scorecard-action/pull/961/files.

Thanks, I'll get a PR out soon and a fix.

naveensrinivasan:
Fixed v2.0.6
```
docker pull gcr.io/openssf/scorecard-action:v2.0.6
v2.0.6: Pulling from openssf/scorecard-action
bd159e379b3b: Pull complete
8bcf04acd839: Pull complete
29b14408ebd9: Pull complete
e908b0653632: Pull complete
[...]

PR #982, opened 2022-10-19
:seedling: Distroless container

naveensrinivasan:
- Distroless containers

This should wait until we have #981 merged 
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,17 +35,6 @@ ARG TARGETOS
 ARG TARGETARCH
 RUN CGO_ENABLED=0 make build
 
-# TODO: use distroless:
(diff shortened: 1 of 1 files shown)


spencerschrock (approved):

spencerschrock:
> Do we have scorecardaction tests running on the `main` tag? This is to avoid #978

This got brought up in the sync, but worth writing it out. We have tests that run every 24 hours and the v2.0.5 issue occurred and was resolved in between two runs: [run 1](https://github.com/ossf-tests/scorecard-action/actions/runs/3270206914) and [run 2](https://github.com/ossf-tests/scorecard-action/actions/runs/3278398635)

Issue #1003, opened 2022-11-01
Use of classic access tokens should be discouraged


jku:
Current installation instructions include this (leaving out some unrelated details):

> 1. Create a Personal Access Token with the following read permissions
>    *  repo > public_repo 
> 2. Copy the token value.
> 3. Create a new repository secret with the following settings:
>    * Name: SCORECARD_READ_TOKEN
>    * Value: the value of the token created in step 1 above.

There are a few issues:
* the token is characterized as a "read" token several times: I believe this is incorrect and saying so creates a false sense of security
* classic tokens are problematic for multiple reasons and should be avoided
* repository secrets are not an appropriate place to store high value secrets

I suggest that  the project
* stops recommending use of classic PATs, and stops describing them as "read" tokens
* recommends that users delete the PATs they have already created for scorecard-action
* (optional) documents how to use fine grained tokens for this purpose

cc @laurentsimon @olivekl

laurentsimon:
That would be great, yes. @jku has volunteered to send a PR for the Action repo. Or do you want to take a stab at it yourself?

jku:
> @jku has volunteered to send a PR for the Action repo

sorry that's a misunderstanding: I would prefer not to do it at least in the next few days. I may be able to have a look later if no-one's taken it.

olivekl:
I would prefer to review the Action update; I don't have the detailed knowledge to confidently rewrite it fully. Happy to set up time to review the remaining token docs together, though. I've been meaning to do a full review of the Scorecards docs for maintenance anyway.

laurentsimon:
I tried using the fine-grained tokens, but they are not supported by graphQL, it seems `Personal access tokens with fine grained access do not support the GraphQL API`, it seems :/

PR #1018, opened 2022-11-22
doc: call out the PAT having write permissions

laurentsimon:
closes https://github.com/ossf/scorecard-action/issues/1003

@jku can you review the changes?


README.md
```diff
@@ -47,24 +47,24 @@ Running the Scorecard action on a fork repository is not supported.
 
 Private repositories need a Personal Access Token (PAT).
 
-Public repositories need a PAT to enable the [Branch-Protection](https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection) check. Without a PAT, Scorecards will run all checks except the Branch-Protection check.
+Public repositories need a PAT to enable the [Branch-Protection](https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection) check. Without a PAT, Scorecards will run all checks except the Branch-Protection check. Due to a limitation of the GitHub permission model, the PAT needs [write permisison to the repository](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes) through the `repo` scope. Beware that the PAT will be stored as a [GitHub encrypted secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets) and be accessible by all the workflows and maintainers of a repository.
```
olivekl, line 50: `Beware that the PAT will be stored as a [GitHub encrypted secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets) and be accessible by all the workflows and maintainers of a repository.`

Will the implications of this warning be clear to the reader or can we spell them out? Why should they be concerned about the PAT/secret being accessible to other workflows and maintainers? What could happen if misused?

If it's a strong warning, I'd make it stand out more: **Important**: the PAT will be stored...

README.md
```diff
@@ -47,24 +47,24 @@ Running the Scorecard action on a fork repository is not supported.
...
 
 GitHub Enterprise repositories are not supported.
 
 ## Installation
 The Scorecards Action is installed by setting up a workflow on the GitHub UI.
 
-**Private repositories**: Scorecards requires authentication using a Personal Access Token (PAT). So if you install Scorecards on a private repository, you will need to follow the optional Authentication step. If you don't, Scorecards will fail to run.
+**Private repositories**: Scorecards requires authentication using a Personal Access Token (PAT). So if you install Scorecards on a private repository, you will need to follow the optional Authentication step. If you don't, Scorecards will fail to run. Beware that the PAT will be stored as a [GitHub encrypted secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets) and be accessible by all the workflows and maintainers of a repository.
```
olivekl, line 57: Suggested: ```Scorecards requires authentication using a Personal Access Token (PAT). If you install Scorecard on a private repository, you must follow the optional Authentication step for Scorecard to run. ```

olivekl (commented):
LGTM with comments

README.md
laurentsimon: They should be worried about it because it increases the chance of the secret leaking to someone who is not the owner of the token. Can you suggest a wording? I'll add the *Warning:* in all occurrences of the text

README.md
laurentsimon: done, thanks

jku (approved):
I just have one comment on the content:
* the copy in starter-workflows/code-scanning is not updated (the two copies could use merging otherwise too, there seem to be small differences)

This looks like an improvement so I'll "approve" based on that but I have a higher level comment as well, will leave it separately.

jku:
Following is based on my understanding that for typical projects, a PAT is only required for the Branch-Protection check, other functionality is unaffected.

I believe that the risk from putting a personal developer token in GH secrets far, far outweighs the benefit that the Branch-Protection check provides. In my opinion *no-one* should be adding a PAT to secrets for that and the instructions should not advice to do so. If this means Branch-Protection check is not available, then that's a small price to pay for doing the right thing.

I think the most correct call is to not advice users to create PATs in the main README, or at most mention them in a separate section for private repositories (but even in that case I personally question the value proposition of using PATs).

laurentsimon:
> I just have one comment on the content:
> 
> * the copy in starter-workflows/code-scanning is not updated (the two copies could use merging otherwise too, there seem to be small differences)
> 
> This looks like an improvement so I'll "approve" based on that but I have a higher level comment as well, will leave it separately.

Let me remove the starter-workflow. The starter-workflow is this one https://github.com/actions/starter-workflows/pull/1830

laurentsimon:
@naveensrinivasan PTAL

naveensrinivasan (approved):
Thanks

PR #1026, opened 2022-12-08
Make the badge text easier to copy

rajbos:
I have copied that line a couple of times and forgot to remove the spaces in it more than once 🙈 . 
Making it a one line will prevent more users from going through this cycle (there where two spaces, so often two extra commits before it worked!).

Also updated the `org` variable --> `owner`, since you can either have an organization repo or a user repo, so it's common to refer to `owner`.


naveensrinivasan (approved):
Thank you!

PR #1031, opened 2022-12-09
feat: Disable pull_request_target

laurentsimon:
Signed-off-by: laurentsimon <laurentsimon@google.com>

This trigger is dangerous so let's disable it.


laurentsimon:
There does not seem to be any uses of scorecard with this trigger at the moment https://github.com/search?q=pull_request_target++ossf%2Fscorecard-action&type=code

naveensrinivasan (commented):
Can you please add some links as to why this is dangerous?

laurentsimon:
> Can you please add some links as to why this is dangerous?

Done, as part of the error message. PTAL

naveensrinivasan (approved):
Thanks

Issue #1028, opened 2022-12-08
Error uploading score data with v2.0.*


rajbos:
I was adding this action to my repo and all uploads with v2 keep on failing. 
```shell
tlog entry created with index: 8678709
MEYCIQDqgM0We/hZyDOs/8z4xgKLBWtTkgnI52BLyTu89tIHdAIhALVwPK6T6oArxGNlLsLRIjdRwih2o9mhZWQRpxwzTR/M
2022/12/08 19:33:00 error processing signature: http response 500, status: 500 Internal Server Error, error: {"code":500,"message":"something went wrong and we are looking into it."}
```

There for no score is available at all:
![image](https://user-images.githubusercontent.com/6085745/206551185-91019ec4-f74f-400e-8fc0-3df9fb337880.png)

Log to the workflow run: [link](https://github.com/devops-actions/load-runner-info/actions/runs/3651486416/jobs/6168754469).

rajbos:
Seems like this is caused by me following best practices and forking the action to my own org. This is not mentioned in the readme! Using the official action makes it work.

The readme only mentions this restriction for the actions that are used (which I do use from their source):
```
scorecard-action:v2 has a new requirement for the job running the ossf/scorecard-action step. The step running this job must belong to this approved list of GitHub actions:

"actions/checkout"
"actions/upload-artifact"
"github/codeql-action/upload-sarif"
```

So if the requirement also is `ossf/scorecard-action` needs to be the official one, then I suggest adding that to the readme as well (I'm happy to create the PR if this is the case). 

Additionally, if that is the case then I do would like to discuss the handling of this, since indeed this is overly permissive, not following best practices (forking), and not useable in an Enterprise Server environment (self-hosting the API). If so, I'll follow up in a different issue.

laurentsimon:
@rajbos is this the workflow that failed https://github.com/devops-actions/load-runner-info/blob/main/.github/workflows/ossf-analysis.yml?

It seems to be using only the Actions listed above. Is your integration in another workflow?

rajbos:
> 

Yes it is

> @rajbos is this the workflow that failed https://github.com/devops-actions/load-runner-info/blob/main/.github/workflows/ossf-analysis.yml?
> 
> It seems to be using only the Actions listed above. Is your integration in another workflow?

[...]

PR #1044, opened 2022-12-16
Complete the list of required actions

rajbos:
Make the documentation explicit by mentioning that this action also needs to be called from the original location, otherwise it will fail with a non-descriptive error, as discussed in #1028.

closes https://github.com/ossf/scorecard-action/issues/1028

Signed-off-by: Rob Bos <rajbos@users.noreply.github.com>


naveensrinivasan (approved):
Thanks

PR #1054, opened 2022-12-21
🌱 Bump scorecard dependency to v4.10.2 to remove a CODEOWNERS printf statement.

spencerschrock:
Signed-off-by: Spencer Schrock <sschrock@google.com>

Bug report at https://github.com/ossf/scorecard/pull/2557 for details.


Makefile
```diff
@@ -1,5 +1,5 @@
 # NOTE: Keep this in sync with go.mod for ossf/scorecard.
-LDFLAGS=-X sigs.k8s.io/release-utils/version.gitVersion=v4.8.0 -X sigs.k8s.io/release-utils/version.gitCommit=c40859202d739b31fd060ac5b30d17326cd74275 -w -extldflags \"-static\"
+LDFLAGS=-X sigs.k8s.io/release-utils/version.gitVersion=v4.10.2 -X sigs.k8s.io/release-utils/version.gitCommit=376f465c111c39c6a5ad7408e8896cd790cb5219 -w -extldflags \"-static\"
```
laurentsimon, line 2: if we release on a tag, we should be able to grab these values dynamically `GITHUB_REF_NAME` and `GITHUB_SHA`. Or we can use some git comment to read them.

laurentsimon (approved):

Makefile
spencerschrock: Yeah, we can look into that. Given the last update was v4.8.0 it's something that's easily forgotten.

PR #1055, opened 2022-12-21
🌱 Bump docker tag for new release.

spencerschrock:
Signed-off-by: Spencer Schrock <sschrock@google.com>

I've manually tested the action, as the steps in [RELEASE.md](https://github.com/ossf/scorecard-action/blob/7da02bf0d58396bc404a7e5aef3e9b0c24dcb9bc/RELEASE.md#validate-the-action) don't actually test any new changes since the docker image is pinned to a version.

I've tested with this command and confirmed the original bug (https://github.com/ossf/scorecard/pull/2557) is fixed.
```
docker run -e INPUT_REPO_TOKEN="$GITHUB_AUTH_TOKEN" \
    -e INPUT_POLICY_FILE="/policy.yml" \
    -e INPUT_RESULTS_FORMAT="sarif" \
    -e INPUT_RESULTS_FILE="results.sarif" \
    -e INPUT_PUBLISH_RESULTS="false" \
    -e GITHUB_WORKSPACE="/github/workflow" \
    -e GITHUB_REF="refs/heads/main" \
    -e GITHUB_EVENT_NAME="branch_protection_rule" \
    -e GITHUB_EVENT_PATH="/testdata/test.json" \
    -e GITHUB_REPOSITORY="nginxinc/nginx-kubernetes-gateway" \
    -e GITHUB_API_URL="https://api.github.com/" \
    -v $PWD:/github/workflow \
    gcr.io/openssf/scorecard-action:latest
 ```

### DIFF
--- a/action.yaml
+++ b/action.yaml
@@ -53,4 +53,4 @@ branding:
 
 runs:
   using: "docker"
-  image: "docker://gcr.io/openssf/scorecard-action:v2.1.1"
+  image: "docker://gcr.io/openssf/scorecard-action:v2.1.2"

### CODE REVIEW

**Review by laurentsimon — APPROVED** [review]
I think the RELEASE.md needs to move the section on testing to *after* the section on updating the tag. Or we use latest tag instead.

**Comment by spencerschrock** [comment]
> I think the RELEASE.md needs to move the section on testing to _after_ we've updated the tag

Until the release happens, the newly tagged Docker image won't exist. So doing it after the tag but before the release also won't work.

I think we need a test version of `action.yaml` which uses `gcr.io/openssf/scorecard-action:latest`

**Comment by laurentsimon** [comment]
> > I think the RELEASE.md needs to move the section on testing to _after_ we've updated the tag
> 
> Until the release happens, the newly tagged Docker image won't exist. So doing it after the tag but before the release also won't work.

Yeah, I meant after release as well. But too late f it's broken.
 
> I think we need a test version of `action.yaml` which uses `gcr.io/openssf/scorecard-action:latest`

What is `latest` in this case? It still points to an old scorecard release, does it not?

Or we need to cut a `v1.2.3-rc` and use that instead to test before the actual release?

**Comment by spencerschrock** [comment]
> > I think we need a test version of `action.yaml` which uses `gcr.io/openssf/scorecard-action:latest`
> 
> What is `latest` in this case? It still points to an old Action release, does it not?
> 
> Or we need to cut a `v1.2.3-rc` and use that instead to test before the actual release?

So right now RELEASE.md mentions a manual dispatch of `scorecards.yml` which calls our public `action.yaml`

https://github.com/ossf/scorecard-action/blob/e38b1902ae4f44df626f11ba0734b14fb91f8f86/.github/workflows/scorecards.yml#L23-L26

But `ossf/scorecard-action@main` is pinned to a specific docker version, so it will always run an old docker release:
https://github.com/ossf/scorecard-action/blob/e38b1902ae4f44df626f11ba0734b14fb91f8f86/action.yaml#L54-L56

I think we need to change `scorecards.yml` to use a private action which references the `latest` docker image, which will have the changes we're trying to test. We need to [setup a private action](https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action#example-using-a-private-action) which has the following lines
```
runs:
  using: "docker"
  image: "docker://gcr.io/openssf/scorecard-action:latest"
```

Issue #1070, opened 2023-01-13
Implement Scorecard action for dependencies


naveensrinivasan:
Currently, Scorecard has scanned over 1,200,000 repositories, which are available in the API. However, when a GitHub project adds a new dependency, customers don't know about the health of the dependency. This can lead to potential issues in the project.

GitHub provides an API to get differences in dependencies between commits (https://docs.github.com/en/rest/dependency-graph/dependency-review?apiVersion=2022-11-28#get-a-diff-of-the-dependencies-between-commits). By utilizing this API in conjunction with the Scorecard API, we can provide the health of the dependency when there is a pull request to the repository. This will help consumers be aware of the health of the dependency and make informed decisions.

### Proposed solution:

- Implement an action that utilizes the GitHub API to get the differences in dependencies between commits.
- Use the Scorecard API to check the health of the new dependencies.
- Display the results of the health check in the pull request.

### Benefits:

- Helps customers make informed decisions about new dependencies
- Increases transparency and trust in the project
- It can potentially prevent issues caused by unhealthy dependencies.

### Additional Information:

The reason for using the API https://docs.github.com/en/rest/dependency-graph/dependency-review?apiVersion=2022-11-28#get-a-diff-of-the-dependencies-between-commits instead of using other OSS projects for getting dependencies is that GitHub provides a mechanism to map dependencies to a GitHub project, which is its unique feature. This is a more reliable way to identify dependencies as it links the dependency to its source. Using this API, we can ensure that we get the correct information about the dependencies and their health. 

It could be something like this https://github.com/naveensrinivasan/scorecard-action/pull/1#issuecomment-1380873657

![image](https://user-images.githubusercontent.com/172697/212215958-e3f23fd2-6b4b-4129-8857-2dcc940a0734.png)

naveensrinivasan:
I am planning to work on this.

laurentsimon:
related to API issues: https://github.com/ossf/scorecard-action/issues/1071

laurentsimon:
IIRC, some of the things to consider:
1. Where do we implement this:
  a. In this Action: 
   - pros: all existing users get it for free if they update the trigger
   - cons: we have rate limit problems. Is that only a problem for our repos or other people also hit this problem? What that number of PR / commits per hour (?) where we hit this problem in practice? Can anything be done in Scorecard to remediate this problem? NOTE: scorecard runs only file-based checks on PR for SARIF results, but *all* the checks for the JSON results uploaded to the API server.
[...]

PR #1095, opened 2023-02-23
Add Dependency Analysis Action and Dockerfile

naveensrinivasan:
New scorecard action https://github.com/ossf/scorecard-action/issues/1070

- Add workflow to publish dependency analysis Docker image
- Add a new filter function to filter slices
- Add a GetScorecardChecks function to get scorecard checks
- Add a GetScore function to get score of a repo
- Add a Validate function to validate token, owner, repo, commitSHA, and PR
- Add a new action file for OSSF Scorecard dependency analysis

https://github.com/ossf-tests/vulpy/pull/2 - python
https://github.com/ossf-tests/vulpy/pull/1 - GitHub action
https://github.com/naveensrinivasan/scorecard-action/pull/23 - golang

Example of the GitHub action https://github.com/ossf-tests/vulpy/blob/master/.github/workflows/scorecard-dependency.yml


dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+				sb.WriteString("<tr>\n")
+				sb.WriteString(fmt.Sprintf("<td>%s</td>\n", v.Severity))
+				sb.WriteString(fmt.Sprintf("<td>%s</td>\n", v.AdvisoryGHSAId))
+				sb.WriteString(fmt.Sprintf("<td>%s</td>\n", v.AdvisorySummary))
+				sb.WriteString(fmt.Sprintf("<td>%s</td>\n", v.AdvisoryURL))
+			}
+			sb.WriteString("</table>\n")
+			sb.WriteString("</details>\n")
+			vulnerabilities += sb.String()
+			continue
```
spencerschrock, line 95: having a vulnerability will skip all the other checks for a repository? I know having a vuln will cause the action to fail, but is it not worth having the other output?

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+					return true
+				}
+			}
+			return false
+		})
+		scorecard.Vulnerabilities = i.Vulnerabilities
+		result += GitHubIssueComment(&scorecard)
+	}
+	// convert pr to int
+	prInt, err := strconv.Atoi(pr)
```
spencerschrock, line 109: this should be moved either to `Validate` or at least before the API calls

Dockerfile-dependency-analysis
```diff
@@ -0,0 +1,35 @@
+# Copyright 2023 Security Scorecard Authors
```
raghavkaul, line 1: Nit: Rename to Dependency-analysis.dockerfile so that syntax highlighting works

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
...
+## Usage
+The OpenSSF Dependency Analysis is a GitHub Action that can be easily incorporated into a workflow. 
+The workflow can be triggered on a pull request event. 
+The action will run on the latest commit on the default branch of the repository, and will create a comment on the pull request with the results of the analysis. 
+An example of the comment can be found [here](https://github.com/ossf-tests/vulpy/pull/2#issuecomment-1442310469).
+
+## Prerequisites
+The actions require enabling the [GitHub Dependency](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review) for the repository.
+
+### Configuration
```
raghavkaul, line 14: Does it make sense to allow users to specify skipped dependencies? Totally fine if that feature is not in this PR.

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Run dependency analysis
+        uses: github.com/ossf/scorecard-action/dependency-analysis@main # Replace with the latest release version.
```
raghavkaul, line 57: Nit: github.com/ needed?

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+	}
+	return &filteredData, nil
+}
+
+// GetGitHubDependencyDiff returns the dependency diff between two commits.
+// It returns an error if the dependency graph is not enabled.
+func GetGitHubDependencyDiff(owner, repo, token, base, head string) (*http.Response, error) {
+	req, err := http.NewRequest("GET",
+		fmt.Sprintf("https://api.github.com/repos/%s/%s/dependency-graph/compare/%s...%s", owner, repo, base, head), nil)
```
raghavkaul, line 276: Can the GitHub client from `go-github` be used?

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	m := make(map[string]DependencyDiff)
+	for _, dep := range *data { //nolint:gocritic
+		m[dep.SourceRepositoryURL] = dep
+	}
+
+	for k, i := range m { //nolint:gocritic
+		url := strings.TrimPrefix(k, "https://")
+		scorecard, err := GetScore(url)
```
raghavkaul, line 74: Should this be a URL + version? Or is there any other way we can make it clear we're running on Git HEAD rather than the version of a package used.

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+		if err != nil && len(i.Vulnerabilities) > 0 {
```
raghavkaul, line 75: Should `err != nil` be handled separately?

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+				sb.WriteString(fmt.Sprintf("<td>%s</td>\n", v.AdvisoryGHSAId))
+				sb.WriteString(fmt.Sprintf("<td>%s</td>\n", v.AdvisorySummary))
+				sb.WriteString(fmt.Sprintf("<td>%s</td>\n", v.AdvisoryURL))
+			}
+			sb.WriteString("</table>\n")
+			sb.WriteString("</details>\n")
+			vulnerabilities += sb.String()
+			continue
+		}
+		scorecard.Checks = filter(scorecard.Checks, func(check Check) bool {
```
raghavkaul, line 97: Does `GetScore` on L74 run all checks? If it does, it seems like we throw away some of its work. Maybe we should filter these checks earlier

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+		return "", fmt.Errorf("failed to get repository: %w", err)
+	}
+
+	return repository.GetDefaultBranch(), nil
+}
+
+// Validate validates the input parameters.
+func Validate(token, repo, commitSHA, pr string) error {
+	if token == "" {
+		return fmt.Errorf("token is empty") //nolint:goerr113
```
raghavkaul, line 147: Nit: Shouldn't ignore lint, would be easier to create a new error type here.

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+	prInt, err := strconv.Atoi(pr)
+	if err != nil {
+		log.Fatal(err)
+	}
+	// create or update comment
+	if vulnerabilities == "" && result == "" {
+		return
+	}
+	if err := createOrUpdateComment(
+		client, owner, ghUser, repo, prInt, "## Scorecard Results</br>\n"+vulnerabilities+"</br>"+result); err != nil {
```
raghavkaul, line 118: ghUser is `GITHUB_ACTOR` which is the user who triggered the workflow - what do we expect GITHUB_ACTOR to be usually? if the workflow is re-triggered manually, ghUser would be the identity of a human repo maintainer - and then we'd try to update their first comment, it seems?

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+	sb.WriteString("</details>")
+	return sb.String()
+}
+
+// GetDependencyDiff returns the dependency diff between two commits.
+// It returns an error if the dependency graph is not enabled.
+func GetDependencyDiff(owner, repo, token, base, head string) (*[]DependencyDiff, error) {
+	message := "failed to get dependency diff, please enable dependency graph https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/configuring-the-dependency-graph" //nolint:lll
+	if owner == "" {
+		return nil, fmt.Errorf("owner is required") //nolint:goerr113
```
raghavkaul, line 235: Nit: Should create an error type.

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+			n = append(n, e)
+		}
+	}
+	return n
+}
+
+// GetScorecardChecks returns the list of checks to run.
+// This uses the SCORECARD_CHECKS environment variable to get the path to the checks list.
+func GetScorecardChecks() ([]string, error) {
+	fileName := os.Getenv("SCORECARD_CHECKS")
```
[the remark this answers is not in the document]
raghavkaul: Agree that it should be parameterized for testability; also wonder if filename is the right approach? If the user is configuring the DependencyDiff action, can their string environment variable be the comma-separated list of checks instead of the name of a file with the list of checks?

raghavkaul (changes requested):
Thanks for working on this, some suggestions for code cleanup.

dependency-analysis/action.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+# Action syntax: https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions.
+
+name: "OSSF Scorecard dependency analysis"
+description: "Run OSSF Scorecard dependency analysis on your repository to get quality metrics on your dependencies."
+author: "OSSF - github.com/ossf/scorecard"
+
+runs:
+  using: "docker"
+  image: "docker://ghcr.io/ossf/scorecard-action-dependency-analysis:latest"
```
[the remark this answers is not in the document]
naveensrinivasan: I want to merge this in and associate it with the pinned SHA.

.github/workflows/publish-dependency-image.yml
```diff
@@ -0,0 +1,45 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+
+env:
+  REGISTRY: ghcr.io
```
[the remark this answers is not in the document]
naveensrinivasan: We can do that. Probably after this PR gets merged in then, we can update it. I would like to get this merged so that we can beta-test it. I agree before going PROD we will update it.

dependency-analysis/main.go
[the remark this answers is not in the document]
naveensrinivasan: > would it make sense to move filename to a paremeter?
> That would centralize all the environment variable reading in `main()` and would be easier to test this function.

I can move the ENV variable to main. 
>also wonder if filename is the right approach? If the user is configuring the DependencyDiff action, can their string environment variable be the comma-separated list of checks instead of the name of a file with the list of checks?

The reason for the file is to make is cleaner to configure instead of passing `19` checks as an argument in the command line. 

Also when we have the policy it is also going to be a file and that was my thought process

dependency-analysis/main.go
naveensrinivasan: It will continue only when it is an error `GetScorecardResult`.

dependency-analysis/README.md
naveensrinivasan: Yes, I agree. That is for later.

dependency-analysis/main.go
naveensrinivasan: The go-github doesn't provide this API, and that was the reason for me to use this.

dependency-analysis/main.go
naveensrinivasan: We cannot pass the version as we have only commit SHA in Scorecard API.

dependency-analysis/main.go
naveensrinivasan: The`GITHUB_ACTOR` https://docs.github.com/en/actions/learn-github-actions/variables would be different if a human runs it, so we won't be updating the existing comment but will be creating new one.

Do you have other thoughts?

dependency-analysis/main.go
naveensrinivasan: For now, can we skip this please?

dependency-analysis/main.go
naveensrinivasan: Sorry, I don't get your point. I am not re-running any check other than filtering it here.

dependency-analysis/README.md
```diff
@@ -0,0 +1,16 @@
+# OpenSSF Dependency Analysis
```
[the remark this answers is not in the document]
naveensrinivasan: 👍 , Fixed it. Thanks

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
...
+This repository contains the source code for the OpenSSF Dependency Analysis project. The aim of the project is to check the security posture of a project's dependencies using the [GitHub Dependency Graph API](https://docs.github.com/en/rest/dependency-graph/dependency-review?apiVersion=2022-11-28#get-a-diff-of-the-dependencies-between-commits) and the [Security Scorecards API](https://api.securityscorecards.dev).
+
+## Usage
+The OpenSSF Dependency Analysis is a GitHub Action that can be easily incorporated into a workflow. 
+The workflow can be triggered on a pull request event. 
+The action will run on the latest commit on the default branch of the repository, and will create a comment on the pull request with the results of the analysis. 
+An example of the comment can be found [here](https://github.com/ossf-tests/vulpy/pull/2#issuecomment-1442310469).
+
+## Prerequisites
+The actions require enabling the [GitHub Dependency](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review) for the repository.
```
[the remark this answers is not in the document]
naveensrinivasan: 👍

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	defaultBranch, err := getDefaultBranch(owner, repo, token)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
+	tc := oauth2.NewClient(context.Background(), ts)
+	client := github.NewClient(tc)
```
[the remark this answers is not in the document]
naveensrinivasan: Agreed, Thanks.

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+	data, err := GetDependencyDiff(owner, repo, token, defaultBranch, commitSHA)
```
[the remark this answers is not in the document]
naveensrinivasan: 👍

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+		sb.WriteString("</table>\n")
+	}
+
+	sb.WriteString("</details>")
+	return sb.String()
+}
+
+// GetDependencyDiff returns the dependency diff between two commits.
+// It returns an error if the dependency graph is not enabled.
+func GetDependencyDiff(owner, repo, token, base, head string) (*[]DependencyDiff, error) {
```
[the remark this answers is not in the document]
naveensrinivasan: 👍

dependency-analysis/main.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	var data []DependencyDiff
+	err = json.NewDecoder(resp.Body).Decode(&data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode dependency diff: %w", err)
+	}
+	// filter out the dependencies that are not added
+	var filteredData []DependencyDiff
+	for _, dep := range data { //nolint:gocritic
```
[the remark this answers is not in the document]
naveensrinivasan: Can we skip this for now?

dependency-analysis/main.go
naveensrinivasan: 👍

dependency-analysis/README.md
naveensrinivasan: 👍

dependency-analysis/main.go
naveensrinivasan: 👍

dependency-analysis/main.go
raghavkaul: I mean the dependency version - if we're running scorecard on the dependency, is it possible to indicate which version of the dependency we're scoring?

dependency-analysis/main.go
raghavkaul: Yes, but can we apply the filter function earlier so that we don't run scorecard checks that we'll filter out anyway?

dependency-analysis/main.go
raghavkaul: If we use a hardcoded username (like "scorecard-dependencydiff" for example) instead of `GITHUB_ACTOR`, it would avoid the issue where DependencyDiff attempts to post a comment with the name of the user who triggered the action.

dependency-analysis/main.go
naveensrinivasan: Again this is an API call to api.securityscorecards.dev, so it shouldn't be an issue as we are just fetching the results for the dependency.

dependency-analysis/main.go
naveensrinivasan: My logic searches for GitHub comments by the USER, and AFAIK, I cannot randomly assign a user that would create a comment with that user as a reference. So the comment would be created by the `GITHUB_ACTOR`. Unless you think it is possible?

dependency-analysis/main.go
raghavkaul: Got it, makes sense.

azeemshaikh38:
Added a question/comment on this https://github.com/ossf/scorecard-action/issues/1070#issuecomment-1448404581.

azeemshaikh38 (changes requested):
Question/comment on this design: https://github.com/ossf/scorecard-action/issues/1070#issuecomment-1448404581

dependency-analysis/main.go
raghavkaul: Ah I just found the docs that mention the difference between `actor` (which stays the same when your workflow is rerun manually) and `triggering_actor`. This is fine.

raghavkaul (approved):
Thanks, LGTM.

naveensrinivasan:
> Question/comment on this design: [#1070 (comment)](https://github.com/ossf/scorecard-action/issues/1070#issuecomment-1448404581)

We are continuing the discussion on the ☝️ thread

naveensrinivasan:
> Question/comment on this design: [#1070 (comment)](https://github.com/ossf/scorecard-action/issues/1070#issuecomment-1448404581)

A friendly ping

dependency-analysis/main.go
```diff
@@ -0,0 +1,357 @@
...
+	m := make(map[string]DependencyDiff)
+	for _, dep := range diff { //nolint:gocritic
+		m[dep.SourceRepositoryURL] = dep
+	}
+
+	for k, i := range m { //nolint:gocritic
+		url := strings.TrimPrefix(k, "https://")
+		scorecard, err := GetScorecardResult(url)
+		if err != nil {
+			if len(i.Vulnerabilities) > 0 {
```
laurentsimon, line 85: Why are we treating Vulns as a special case? 

I worry our data is inaccurate, ie we're using a dependency HEAD whereas a user is installing a particular package version. There are other tools that will do a better job at this than Scorecard, like dependabot.

I'm thinking we may even remove the vuln results entirely because of how poor our data may be. Otherwise users will be confused and unhappy that we're claiming there are vulns in a package if there are none.

dependency-analysis/main.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	return nil
+}
+
+// GitHubIssueComment returns a markdown string for a GitHub issue comment.
+func GitHubIssueComment(checks *ScorecardResult) string {
+	if checks.Repo.Name == "" {
+		return ""
+	}
+	sb := strings.Builder{}
```
laurentsimon, line 214: maybe create a template file rather than hardcoded this layout in code. It' pretty hard to read otherwise

dependency-analysis/main.go
```diff
@@ -0,0 +1,357 @@
...
+	return n
+}
+
+// GetScorecardChecks returns the list of checks to run.
+// This uses the SCORECARD_CHECKS environment variable to get the path to the checks list.
+func GetScorecardChecks(fileName string) ([]string, error) {
+	if fileName == "" {
+		// default to critical and high severity checks
+		return []string{
+			"Dangerous-Workflow",
```
laurentsimon, line 317: Should we instead be returning just the aggregate score and link to deps.dev for better UX? Or at least provide a link to deps.dev for users to see more results?

dependency-analysis/main_test.go
```diff
@@ -0,0 +1,223 @@
...
+			fileName := ""
+			if tt.fileContent != "" {
+				fileName = path.Join(dir, "scorecard.txt")
+			}
+			got, err := GetScorecardChecks(fileName)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("GetScorecardChecks() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
laurentsimon, line 101: nit: use [go-cmp](https://github.com/google/go-cmp) package for deep comparison

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
+# OpenSSF Scorecard Dependency Analysis
+
+This repository contains the source code for the OpenSSF Dependency Analysis project. The aim of the project is to check the security posture of a project's dependencies using the [GitHub Dependency Graph API](https://docs.github.com/en/rest/dependency-graph/dependency-review?apiVersion=2022-11-28#get-a-diff-of-the-dependencies-between-commits) and the [Security Scorecards API](https://api.securityscorecards.dev).
+
+## Usage
+The OpenSSF Dependency Analysis is a GitHub Action that can be easily incorporated into a workflow. 
+The workflow can be triggered on a pull request event. 
+The action will run on the latest commit on the default branch of the repository, and will create a comment on the pull request with the results of the analysis. 
+An example of the comment can be found [here](https://github.com/ossf-tests/vulpy/pull/2#issuecomment-1442310469).
```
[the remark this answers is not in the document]
naveensrinivasan: >are we showing the aggregate score in the summary for each result

We aren't showing aggregate score. Do you think we should show the aggregate score?

>Are we creating a new comment for each run?

We will not create results for reach run. Only if the user manually runs it only then we create another comment.

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Prerequisites
+The actions require enabling the [GitHub Dependency Review](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-dependency-review) for the repository.
```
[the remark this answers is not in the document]
naveensrinivasan: > Is this really true? For go projects, unlikely. For other projects, is it? We don't need an exact dependency tag, so it should not be needed. Does GitHub API return dependency name / repo without enabling this feature?

I have tested for go projects and here is python package https://github.com/ossf-tests/vulpy/pull/2/files. For teh GH API to return results it needs this feature to be tuned on.

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Configuration
+The action can be configured using the following inputs:
+
+- `SCORECARD_CHECKS`: This environment variable takes a file containing a list of checks to run.
```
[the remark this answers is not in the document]
naveensrinivasan: We can give users the option to exclude certain repositories or checks for a particular repo, which we can make possible with the file option. 

This will also give us the ability to extend the feature in the future. That's the idea.

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
...
+### Installation
+The action can be installed by adding the following snippet to the workflow file:
+```yaml
+name: scorecard-dependency-analysis
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+permissions:
+  pull-requests: write # Required to create a comment on the pull request.
```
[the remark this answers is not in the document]
naveensrinivasan: I haven't tested with a private repo. I will test it and update it.

dependency-analysis/action.yaml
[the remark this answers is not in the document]
naveensrinivasan: Yes it is till we cut a release. 👍

dependency-analysis/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+jobs:
+  dependency-analysis:
+    name: Scorecards dependency analysis
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_PR_NUMBER: ${{ github.event.number }}
```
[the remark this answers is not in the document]
naveensrinivasan: ${{ github.event.number }} isn't directly accessible within the GH action.

dependency-analysis/main.go
```diff
@@ -0,0 +1,357 @@
...
+		log.Fatal("invalid repo")
+	}
+	owner := ownerRepo[0]
+	repo = ownerRepo[1]
+	checks, err := GetScorecardChecks(fileName)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
```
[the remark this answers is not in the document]
naveensrinivasan: This is GH Client access.
```
ts := oauth2.StaticTokenSource(&oauth2.Token{AccessToken: token})
	tc := oauth2.NewClient(context.Background(), ts)
	client := github.NewClient(tc)
```	

I think from the code we should be able to make it being used in GH Client API. Unless you think it is still necessary.

dependency-analysis/main.go
naveensrinivasan: OK. I agree. I can remove the Vulns.

dependency-analysis/main.go
naveensrinivasan: I agree. I can do that.

dependency-analysis/main.go
naveensrinivasan: We can add the aggregate score but not every repo has the deps.dev and that is the reason not to include deps.dev.

dependency-analysis/README.md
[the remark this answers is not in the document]
laurentsimon: We should have a config file in this case then. Standard way to pass parameters to an Action is not env variables. It's inputs.

dependency-analysis/main.go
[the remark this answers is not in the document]
laurentsimon: auth token must bee needed to use the GITHUB_TOKEN. That makes sense now :-)

dependency-analysis/README.md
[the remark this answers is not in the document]
laurentsimon: Very interesting... and unexpected to me

azeemshaikh38 (approved):

dependency-analysis/README.md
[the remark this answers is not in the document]
laurentsimon: so we run it once for the first commit, and then only if the user asks to run?
Maybe add to the comment description to explain to users that they can re-run to update?

spencerschrock:
Closing as this was implemented in a separate repo instead.
https://github.com/ossf/scorecard-dependencyanalysis

PR #1111, opened 2023-03-24
🌱 Bump github.com/ossf/scorecard/v4 from 4.10.2 to 4.10.5

spencerschrock:
In preparation for a new scorecard-action release. Will let the e2e test run for a few days after this PR.


Makefile
```diff
@@ -1,5 +1,5 @@
 # NOTE: Keep this in sync with go.mod for ossf/scorecard.
-LDFLAGS=-X sigs.k8s.io/release-utils/version.gitVersion=v4.10.2 -X sigs.k8s.io/release-utils/version.gitCommit=376f465c111c39c6a5ad7408e8896cd790cb5219 -w -extldflags \"-static\"
```
laurentsimon, line 2: can we compute this dynamically?

laurentsimon (approved):

Makefile
spencerschrock: We can get the version from the `go.mod` pretty easily I would guess. Not sure about the commit though

PR #1169, opened 2023-06-12
:sparkles: Send rekor tlog index to webapp when publishing results

spencerschrock:
Related:
https://github.com/ossf/scorecard-webapp/issues/135#issuecomment-1512120824 
https://github.com/ossf/scorecard-webapp/pull/412

`SignBlobCmd` unfortunately doesn't expose the tlog index directly (although it does print it stdout). So this PR uses the `--bundle` option of cosign to write it to a temp file, then parse the bundle to get the tlog index.


signing/signing.go
```diff
@@ -122,10 +140,12 @@ func (s *Signing) ProcessSignature(jsonPayload []byte, repoName, repoRef string)
 		Result      string `json:"result"`
 		Branch      string `json:"branch"`
 		AccessToken string `json:"accessToken"`
+		TlogIndex   int64  `json:"tlogIndex"`
```
raghavkaul, line 143: This is `logIndex` in cosign.bundle but `tlogIndex` here, is that intentional?

spencerschrock: I tried to provide additional context as a POST field. When reading `logIndex` in `cosign.bundle`, it's easy to know which log index it's referring to. When sending it to the webapp, there's no cosign context clue, so I had added `tlogIndex`.

raghavkaul (approved):

PR #1221, opened 2023-07-07
:book: Add package comments.

spencerschrock:
The linter (when run on commits to main after merge) has been failing due to these for a while now. I took my best guess on a short description for these.

### DIFF
--- a/install/install.go
+++ b/install/install.go
@@ -14,6 +14,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+// Package install contains functionality to install the OpenSSF Scorecard workflow
 package install
 
 import (
--- a/install/install.go
+++ b/install/install.go
@@ -14,7 +14,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-// Package install contains functionality to install the OpenSSF Scorecard workflow
+// Package install contains functionality to install the OpenSSF Scorecard workflow.
 package install
 
 import (
--- a/entrypoint/entrypoint.go
+++ b/entrypoint/entrypoint.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package entrypoint provides functionality to run Scorecard.
 package entrypoint
 
 import (
--- a/github/github.go
+++ b/github/github.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Package github provides repo information from GitHub.
 package github
 
 import (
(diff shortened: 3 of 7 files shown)


justaugustus (changes requested):
> The linter (when run on commits to main after merge) has been failing due to these for a while now.

@spencerschrock -- Before merging this, can you add a commit to ensure the linter fails on PRs. The failure case you've mentioned suggests that how the linter runs in PR CI or locally is failing to catch the lint failures.

Can you push a commit that drops this config line and see what happens?
https://github.com/ossf/scorecard-action/blob/33bb591983cdd3feffdbe87b9719c784f7e90ac1/.github/workflows/golangci.yml#L26

If the existing issues are too many to tackle easily, can you add a TODO comment to that config line and a tracking issue for us to handle in a follow-up?

spencerschrock:
> > The linter (when run on commits to main after merge) has been failing due to these for a while now.
> 
> @spencerschrock -- Before merging this, can you add a commit to ensure the linter fails on PRs. The failure case you've mentioned suggests that how the linter runs in PR CI or locally is failing to catch the lint failures.

The linter does catch things on PRs, but `only-new-issues` doesn't work on `push` events. 


> If the existing issues are too many to tackle easily

I believe these are the only ones, as I checked with `golangci-lint run -c .golangci.yml` when resolving these

spencerschrock:
> Can you push a commit that drops this config line and see what happens?

Assuming we've gotten all of the lint issues. Dropping the `only-new-issues` option shouldn't matter. I'm just wondering if it makes a difference when the linter updates and starts checking (and complaining) for things it wasn't before.

laurentsimon (approved):

spencerschrock:
@justaugustus friendly ping, I believe all the current issues in the codebase are fixed. 

I'm leaning towards keeping `only-new-issues` so unrelated PRs don't need to address linter issues that crop up in other code.

install/install.go
```diff
@@ -14,6 +14,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+// Package install contains functionality to install the OpenSSF Scorecard workflow
```
justaugustus, line 17:
```suggestion
// Package install contains functionality to install the OpenSSF Scorecard workflow.
```

Linter should flag that package comments should have end with a period.

justaugustus (changes requested):
> I'm leaning towards keeping only-new-issues so unrelated PRs don't need to address linter issues that crop up in other code.

@spencerschrock -- sounds good!
One fix before merge!

justaugustus (approved):

PR #1226, opened 2023-07-10
Authenticate get repo call if authentication info available

rajbos:
Building the missing authentication for private repos and running against EMU / GHES environments. Setup is straightforward: if there is an authentication token available (which is used inside of `scorecard`, then we can authenticate the call to get the repository information here as well.

### DIFF
--- a/github/github.go
+++ b/github/github.go
@@ -101,6 +101,13 @@ func (c *Client) ParseFromURL(baseRepoURL, repoName string) (RepoInfo, error) {
 		return ret, fmt.Errorf("error creating request: %w", err)
 	}
 
+	// authenticate the request if there is a token
+	// this will lower the change of hitting the rate limit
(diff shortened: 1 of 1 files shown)


github/github.go
```diff
@@ -101,6 +101,13 @@ func (c *Client) ParseFromURL(baseRepoURL, repoName string) (RepoInfo, error) {
 		return ret, fmt.Errorf("error creating request: %w", err)
 	}
 
+	// authenticate the request if there is a token
+	// this will lower the change of hitting the rate limit
+	auth, present := os.LookupEnv("GITHUB_AUTH_TOKEN")
+	if present {
+		req.SetBasicAuth("x", auth)
+	}
+
```
spencerschrock, line 110: `Client` already uses the roundtripper from scorecard that will read this environment variable. 
What do you think about changing the call below to use a http.Client that uses that roundtripper?

PR #1229, opened 2023-07-11
Update access rights needed to see the results in code scanning

rajbos:
GitHub has changed the visibility of the alerts to every contributor with `write` access instead of only admins. `maintainers` as previously mentioned in the text is a different security role. More info [in the docs here](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/managing-code-scanning-alerts-for-your-repository).

This PR updates the wording so it is correct again.


naveensrinivasan (approved):

naveensrinivasan:
Thanks

PR #1258, opened 2023-09-11
Fix typo in token input docs

aabouzaid:
It's trivial but annoying because the workflow doesn't fail when using `repo-token` but shows a warning in the annotation section.

```
Unexpected input(s) 'repo-token', valid inputs are ['entryPoint', 'args', 'results_file', 'results_format', 'repo_token', 'publish_results', 'internal_publish_base_url', 'internal_default_token']
```


spencerschrock (approved):
Thanks!

PR #1279, opened 2023-10-13
:book: Move token discussion out of main README.

spencerschrock:
Now that repo rules work with the default GITHUB_TOKEN, there's little need to recommend them.


docs/authentication/classic-token.md
```diff
@@ -0,0 +1,15 @@
+# "Classic" Personal Access Token (PAT) Requirements and Risks
+Certain features require a Personal Access Token (PAT).
+We recommend you use a fine-grained token as described in [Authentication](/docs/authentication/fine-grained-auth-token.md).
```
pnacht, line 3:
```suggestion
We recommend you use a fine-grained token as described in [Authentication with Fine-grained PAT](/docs/authentication/fine-grained-auth-token.md).
```

or "as described [here]" to avoid repetition.

"in [Authentication]" seems odd given that this file is also in docs/authentication.

docs/authentication/fine-grained-auth-token.md
```diff
@@ -0,0 +1,35 @@
+# Authentication with Fine-grained PAT (optional)
+
+For repositories that want to detect their classic Branch Protection rules, or webhooks, we suggest you create a fine-grained Personal Access Token (PAT) that Scorecard may use for authentication.
```
laurentsimon, line 3: does the default `GITHUB_TOKEN` work for webhook too?
Does the sentence `Scorecard can run successfully with the workflow's default `GITHUB_TOKEN`, which is our recommended approach.` not apply to webhooks then?

spencerschrock: no, it doesn't work for webhook and will always need an admin read token.
But webhook is experimental and not accessible from Scorecard Action anyway

laurentsimon (approved):

PR #1290, opened 2023-11-03
Use versions instead of commits

OskarStark:
Or is there a specific reason for the commit hashes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,12 +192,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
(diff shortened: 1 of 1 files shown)


spencerschrock:
> Or is there a specific reason for the commit hashes?

It's intentional, for security and stability reasons. Here are some resources where we/others try to explain our rationale.

https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
https://github.com/sethvargo/ratchet#problem-statement

That being said, the versions/SHAs in are README are out of date and could use an update. If you'd like to update them I'd be happy to review the contribution, otherwise feel free to close this PR.

PR #1320, opened 2024-01-10
Spelling

jsoref:
These two commits are quite different.


cmd/installer/main_test.go
```diff
@@ -75,7 +75,7 @@ func createRandomTestRepository(owner string, autoinit bool) (*github.Repository
 		_, resp, err := client.Repositories.Get(context.Background(), owner, repoName)
 		if err != nil {
 			if resp.StatusCode == http.StatusNotFound {
-				// found a non-existent repo, perfect
+				// found a nonexistent repo, perfect
```
jsoref, line 78: happy to drop, but it's an accepted word in US English, so it should be preferred

signing/testdata/results.sarif
```diff
@@ -145,7 +145,7 @@
...
...ils**:\n\nRisk: `High` (possibly vulnerable to attacks on known flaws)  \n\n\n\nThis check tries to determine if the project uses a dependency update tool,\n\nspecifically [dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates) or\n\n[renovatebot](https://docs.renovatebot.com/configuration-options/). Out-of-date\n\ndependencies make a project vulnerable to known flaws and prone to attacks.\n\nThese tools automate the process of updating dependencies by scanning for\n\noutdated or insecure requirements, and opening a pull request to update them if\n\nfound.   \n\n\n\nThis check can determine only whether the dependency update tool is enabled; it\n\ndoes not ensure that the tool is run or that the tool's pull requests are\n\nmerged.   \n\n\n\nNote: A project that fulfills this criterion with other tools may still receive\n\na low score on this test. There are many ways to implement dependency updates,\n\nand it is challenging for an automated tool like Scorecard to detect them all. A\n\nlow score is therefore not a definitive indication that the project is at risk.  \n\n"
```
jsoref, line 148: I can't figure out if this is an input or just a sample output. If it's a sample output, I can't find the origin for it: https://github.com/search?q=org%3Aossf+%2F%5Cbaditional%5Cb%2F&type=code

spencerschrock: This file is both sample output, and test input. It's generated using when the action runs on a repo, and the code which produces that text would be in https://github.com/ossf/scorecard.

But here, it was also being used as input to a test, where we were looking up a record with a hash, so we shouldn't modify this file (except for deleting it when no longer needed).

spencerschrock:
Given the optional developer-facing nature of one fix, and the fact the other shouldn't be modified, I'm going to close this out. Thanks for the interest in the action.

If you want to fix `setup` vs `set up`, feel free to send a PR to https://github.com/ossf/scorecard

PR #1323, opened 2024-01-20
Bump actions in README

### LINKED ISSUE #1287: Change Scorecard Installation Template to use actions/checkout on version 4
When I'm installing Scorecard directly through GitHub, the template for Scorecard workflow is using version 3.1.0, which should be updated since they have already released a new major version.

I tried to investigate where this change should be made but I couldn't figure it out. If you tell me how to do it I can contribute with a PR.

Goooler:
Closes #1287.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        uses: actions/checkout@v4
         with:
           persist-credentials: false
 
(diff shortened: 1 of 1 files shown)


Goooler:
> In the README, this makes it easy for things to go out of date. 

Yeah, I was thinking about this too. I suggest we can use major version in README to let users obtain the latest minor updates.

spencerschrock:
Resolved through #1352. Dependabot will keep the example up-to-date, so we wont need to update it here anymore.

> I suggest we can use major version in README

Since our README example is in our control, we are going to advocate for dependency pinning still. 
https://docs.github.com/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

I did try to switch our [starter workflow](https://github.com/actions/starter-workflows/blob/main/code-scanning/scorecard.yml) to using the major versions since that's annoying to upgrade, but their CONTRIBUTING.md requires the use of hash pinning.

PR #1354, opened 2024-03-25
:seedling: Remove JavaScript CodeQL linting

adamdmharvey:
This PR eliminates the JavaScript CodeQL scanning in the Actions workflows.

Currently, each PR scans both Go and JavaScript:

image

However, the only JavaScript in the repo is a single file.  This file was added via https://github.com/ossf/scorecard-action/pull/23 when the CodeQL workflow was breaking; but it was breaking because there was NO CodeQL scanning language set at the time. (so the "fix" was to go from none, to JavaScript, adding the single line JavaScript file which just spits out to console the word "codeql" 😁 ) But since Go was added, JavaScript should no longer be necessary.

This will eliminate about 1.5 minutes of CodeQL workflow scanning during each actions run.

No release should be necessary.


adamdmharvey:
Here's how it now looks on my fork:

image

spencerschrock:
> But since Go was added, JavaScript should no longer be necessary.

I believe the JS analysis also covers things like GitHub Actions:
https://codeql.github.com/codeql-query-help/javascript/js-actions-command-injection/

adamdmharvey:
> It's caught things for us before on ossf/scorecard

Thanks for the feedback!  Will check into that.

For this repo at least, the Action notes:

> CodeQL scanned 1 out of 1 JavaScript files in this invocation. Check the status page for overall coverage information: https://github.com/ossf/scorecard-action/security/code-scanning/tools/CodeQL/status/

I checked an action on `ossf/scorecard`, and indeed *that* one is scanning 222 "extracted" JavaScript files (even though the repo contains one actual file). Seems it's extracting JavaScript OUT of some of the workflows themselves, which it's not doing in this repo.

adamdmharvey:
Looks like the difference may be this repo isn't doing:

```diff
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@cdcdbb579706841c47f7063dda365e292e5cad7a # v1
       with:
         languages: ${{ matrix.language }}
+        queries: +security-extended
```

Maybe the alternative is to turn that ON here too?

spencerschrock:
At least based on [this doc](https://docs.github.com/en/code-security/code-scanning/managing-your-code-scanning-configuration/javascript-typescript-built-in-queries)

"[Expression injection in Actions](https://codeql.github.com/codeql-query-help/javascript/js-actions-command-injection/)" is listed for both default and `security-extended`

adamdmharvey:
Will give this a close, based on the point re: validating extracted/interpreted JS through embeddings in actions.  Good feedback thx!

PR #1403, opened 2024-07-10
Updated descriptions on GitHub Action install

ghost:
Proactive PR to update the description of the Workflow setup in GitHub action (it looks now different)

P.S Screenshots are in darkmode (saves eye and energy - I can imagine)

PR #1416, opened 2024-07-25
Installation docs update

### LINKED ISSUE #1409: Installation instructions should be updated/improved
After following the[ installation instructions](https://github.com/ossf/scorecard-action#installation), our group found that the following can be updated/improved:

- Outdated screenshots
- Add additional step in instructions (search workflow for OSSF)

Me and @silasvm, @jordin221, @JeremiahAHoward, @klbynum, @emmet0r would love to work on this problem!

JeremiahAHoward:
My team and I updated the installation screenshots, this fixes #1409.

I worked with @emmet0r, @silasVM, @wavyeli32, @jordin221, @klbynum.

Shoutout to @marcinguy for working on #1403 which provided some insight into helping us.


JeremiahAHoward:
What do you think @spencerschrock?

justaugustus:
Thanks for the PR, @JeremiahAHoward!
Assigning @lelia for review.

lelia (changes requested):
Thanks so much for your contribution @JeremiahAHoward ! 

I'll leave some minor phrasing feedback below, and am also going to push a suggested Markdown formatting change to better handle the numbered lists going forward (it's a bit tricky to suggest the proper indentation required for each step, so it's probably easier to just show you the difference in an actual commit). 

It looks like you'll also need to bring the branch up-to-date with the latest from `main`, and we'll need to figure out why Codecov [started failing](https://github.com/ossf/scorecard-action/runs/28030697213) as a result of deleting `install04.png` from the `images` directory.

README.md
```diff
@@ -56,15 +56,21 @@ GitHub Enterprise repositories are not supported.
 
 ![image](/images/install02.png)
 
-3) Then click "Add tool".
+3) Depending on whether or not the repo already has a scanning tool configured, you will see "Add tool" or "Configure scanning tool". If you see "Add tool" click it and skip step 3b. If you see "Configure scanning tool" proceed to step 3b.
```
lelia, line 59: As you'll see in the commit below, I'd suggest using generic references to numbered steps within the text body so that way, if we add more steps in the future, we won't need to revise all the references.

```suggestion
3) Depending on whether the repository already has a scanning tool configured, you will see "Add tool" or "Configure scanning tool". If you see "Add tool" click it and skip step `b`. If you see "Configure scanning tool" proceed to step `b`.
```

README.md
```diff
@@ -56,15 +56,21 @@ GitHub Enterprise repositories are not supported.
...
 
 ![image](/images/install03.png)
+![image](/images/configureScanningToolSS.png)
```
lelia, line 62: General question about the image names — does the `SS` suffix stand for anything in particular?

In looking at the [existing content](https://github.com/ossf/scorecard-action/tree/main/images) stored under `/images/`, it looks like the general pattern is to use all-lowercase image names, so we may want to drop the `SS` and revise these new image names accordingly.

spencerschrock:
> Codecov [started failing](https://github.com/ossf/scorecard-action/runs/28030697213)

Codecov has some rate limiting issues and runs inconsistently. We never migrated to their new format (see https://github.com/ossf/scorecard/issues/3862). Considering this is a README only change, I wouldn't stress on the coverage.

lelia:
@JeremiahAHoward as promised, check out [this commit](https://github.com/lelia/scorecard-action/pull/2/commits/39dcd67018a0b369a5b08289a81fa25c82953cbc) for an example of the [ordered lists](https://www.markdownguide.org/basic-syntax/#ordered-lists) syntax I was referencing above. 

Feel free to use that as inspiration, or copy-paste the raw markdown into your branch 🙂

README.md
spencerschrock: >  if we add more steps in the future, we won't need to revise all the references.

Personally, I think one benefit of markdown is the ability to read it still even when not rendered by GitHub or other platforms. So I'm on the fence about seeing a bunch of repeated 1's, but I don't claim to know markdown best practices.

JeremiahAHoward:
@lelia, @spencerschrock Thank you both for your feedback on our pull request. We addressed the edits asked of us and decided to leave the numbered steps to continue numerically, to preserve markdown readability. We noticed as long as the indentation was properly formatted in markdown, GitHub will automatically correct any discrepancies in list numbering. Please let us know if there is anything else you notice :)

marcinguy (approved):

README.md
lelia: That's fair — it's more of a stylistic preference than a best practice.

lelia (approved):
LGTM

spencerschrock (approved):
Thanks all!

PR #1423, opened 2024-08-09
use Scorecard library entrypoint instead of Cobra hooking

### LINKED ISSUE #1071: feat: run scorecard only once
The Action runs scorecard twice: once for the SARIF results, and once to upload the results to the API server. This increases rate limits and is also slower.

We need to run scorecard once only, then:
1. filter out the results depending on trigger type (PR vs not PR), and generate the SARIF output
2. Use all the results and generate the JSON result to upload to the API server

/cc @naveensrinivasan @azeemsgoogle

spencerschrock:
Scorecard V5 released a new entrypoint, so make use of it instead of
hooking into the underlying Cobra CLI. This gives us more flexibility
when running Scorecard, such as writing the result to multiple formats.

The code got moved around a lot, my hope is it's organized more
clearly now, but I'll try to summarize it for the sake of the reviewer.

`entrypoint/entrypoint.go` used to do several things
1. Read the Scorecard Action options, which now takes place in our `main.go`
https://github.com/ossf/scorecard-action/blob/486e8de2dca09b3f1cfac2bbec765156de6795a1/main.go#L35-L39

2. All sorts of pre/post hooks to redirect output to both stdout and a result file. Which is done explicitly in a new format helper function.
https://github.com/ossf/scorecard-action/blob/486e8de2dca09b3f1cfac2bbec765156de6795a1/internal/scorecard/format.go#L47-L53

`signing/signing.go` re-ran scorecard via `GetJSONScorecardResults`, which is now a simple extra formatting call.

https://github.com/ossf/scorecard-action/blob/486e8de2dca09b3f1cfac2bbec765156de6795a1/main.go#L54-L58

In terms of testing, separating the format code makes it more testable! But the rest of it is very hard to unit test. I'll look into testing the rest of it.

Fixes #1071


spencerschrock:
> In terms of testing, separating the format code makes it more testable! But the rest of it is very hard to unit test. I'll look into testing the rest of it.

Testing the Dockerfile locally, the JSON output is identical (using `scdiff` to compare), and the SARIF matches. The nightly e2e tests would test for fatal errors after merge as well.

But there aren't any great e2e tests currently setup before merge.

internal/scorecard/format.go
```diff
@@ -0,0 +1,88 @@
+// Copyright 2024 Security Scorecard Authors
```
justaugustus, line 1:
```suggestion
// Copyright 2024 OpenSSF Scorecard Authors
```

internal/scorecard/format_test.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2024 Security Scorecard Authors
```
justaugustus, line 1:
```suggestion
// Copyright 2024 OpenSSF Scorecard Authors
```

internal/scorecard/scorecard.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2024 Security Scorecard Authors
```
justaugustus, line 1:
```suggestion
// Copyright 2024 OpenSSF Scorecard Authors
```

main.go
```diff
@@ -31,22 +32,38 @@ func main() {
...
+	if err := scorecard.Format(&result, opts); err != nil {
+		log.Fatal(err)
 	}
 
-	if os.Getenv(options.EnvInputPublishResults) == "true" &&
-		// `pull_request` do not have the necessary `token-id: write` permissions.
-		triggerEventName != "pull_request" {
-		// Get json results by re-running scorecard.
-		jsonPayload, err := signing.GetJSONScorecardResults()
+	// `pull_request` do not have the necessary `token-id: write` permissions.
```
justaugustus, line 50:
```suggestion
	// `pull_request` does not have the necessary `token-id: write` permissions.
```

justaugustus (changes requested):
Minor nits.
Looks much cleaner, @spencerschrock!!

justaugustus (approved):
Approving with code suggestions applied from https://github.com/ossf/scorecard-action/pull/1423#pullrequestreview-2231274945.

PR #1453, opened 2024-10-07
:seedling: publish docker images to GitHub Container Registry

spencerschrock:
The goal is to use GHCR to replace Google Container Registry (GCR) for future Scorecard Action releases to reduce network egress costs. These workflows will build two types of images:

  1. Release images, which are tagged following a v1.2.3 pattern. These container images will be retained indefinitely.
  2. Per-commit images for each push to main. These images are used when testing the action, and will be removed after a week.

The workflow was primarily based on GitHub's [example workflow](https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#publishing-a-package-using-an-action). You can see this working in my fork:

- A normal push to main, generating a `latest` image, with no attestation [workflow](https://github.com/spencerschrock/scorecard-action/actions/runs/11220664575/job/31189323002)
- A tagged image, generating a release image, with an [attestation](https://github.com/spencerschrock/scorecard-action/attestations/2367773). [workflow](https://github.com/spencerschrock/scorecard-action/actions/runs/11220758997/job/31189608920)
- The cleanup workflow, deleting untagged images on [schedule](https://github.com/spencerschrock/scorecard-action/actions/runs/11216103800/job/31174674613), or on [dispatch](https://github.com/spencerschrock/scorecard-action/actions/runs/11188143839/job/31106421523#step:2:12).


.github/workflows/ghcr-retention.yml
```diff
@@ -0,0 +1,22 @@
+name: Delete untagged GHCR images
+on:
+  schedule:
+    - cron:  '17 10 * * TUE' # Tuesday morning at 10:17 UTC
+  workflow_dispatch:
```
raghavkaul, line 5: Necessary?

raghavkaul (commented):
What is the network egress cost and why does it not apply with GHCR?

spencerschrock:
> What is the network egress cost

When someone runs the action, it pulls our gcr.io image, which has [pricing associated](https://cloud.google.com/storage/pricing#network-pricing) with it
> General network usage applies for any data read from your Cloud Storage bucket that does not fall into one of the above categories or the Always Free usage limits. For example, **general network usage applies when data moves from a Cloud Storage bucket to the Internet**.

Monthly Usage | Data transfer to Worldwide Destinations (excluding Asia & Australia)(per GB) | Data transfer to Asia Destinations (excluding China, but including Hong Kong)(per GB) | Data transfer to China Destinations (excluding Hong Kong)(per GB) | Data transfer to Australia Destinations and Data transfer from Cloud Storage regions located in Australia(per GB) | Inbound data transfer
-- | -- | -- | -- | -- | --
0-1 TB | $0.12 | $0.12 | $0.23 | $0.19 | Free
1-10 TB | $0.11 | $0.11 | $0.22 | $0.18 | Free
10+ TB | $0.08 | $0.08 | $0.20 | $0.15 | Free


> why does it not apply with GHCR?

GHCR is [free for public packages](https://docs.github.com/en/billing/managing-billing-for-your-products/managing-billing-for-github-packages/about-billing-for-github-packages#about-billing-for-github-packages). And also has a section later on about GitHub Actions
> All data transferred out, when triggered by GitHub Actions, and data transferred in from any source is free. We determine you are downloading packages using GitHub Actions when you log in to GitHub Packages using a GITHUB_TOKEN.

.github/workflows/ghcr-retention.yml
spencerschrock: I'm not sure which line was supposed to be highlighted, but assuming it's about the `workflow_dispatch` trigger, no it's not technically necessary but being able to manually run a scheduled event seemed reasonable.

raghavkaul (approved):

PR #1507, opened 2025-02-10
limit codeQL push trigger to main branch

spencerschrock:
I noticed the codeQL action running twice on a PR I updated, once because of the pull_request trigger, and once because of the push trigger. This matches what the starter workflow has:

https://github.com/actions/starter-workflows/blob/a00915e13e33549ccc21799315b83d80afa2757c/code-scanning/codeql.yml#L14-L18


justaugustus (approved):
Good catch, @spencerschrock!

PR #1582, opened 2025-09-24
:book: Add missing ENV to development docs

deivid-rodriguez:
Without setting this env, I get the following:

```
$ docker run -e INPUT_REPO_TOKEN="$GITHUB_AUTH_TOKEN" -e INPUT_POLICY_FILE="/policy.yml" -e INPUT_RESULTS_FORMAT="sarif" -e INPUT_RESULTS_FILE="results.sarif" -e INPUT_PUBLISH_RESULTS="false" -e GITHUB_WORKSPACE="/" -e GITHUB_REF="refs/heads/main" -e GITHUB_EVENT_NAME="branch_protection_rule" -e GITHUB_EVENT_PATH="/testdata/fork.json" -e GITHUB_REPOSITORY="ossf/scorecard" -it scorecard-action:testing
2025/09/24 17:27:09 getting repo info from file: /testdata/fork.json
2025/09/24 17:27:09 getting repo info from URL: repos/ossf/scorecard
FromURL: error receiving response: Get "repos/ossf/scorecard": unsupported protocol scheme "".
FromRepoInfo: error receiving response: Get "repos/ossf/scorecard": unsupported protocol scheme "".
2025/09/24 17:27:09 creating new options: parsing repo info: GitHub repo info inaccessible
```

Passing the `GITHUB_API_URL` env as well seems to fix it.

PR #1583, opened 2025-09-24
:book: Fix recommended command to test the image in development

deivid-rodriguez:
Because no testdata files are actually copied to the image, we fail to read the event file and fallback to fetching the information from GitHub API.

Howver, that's not actually working because the `GITHUB_API_URL` environment is missing.

To fix it, we could pass the `GITHUB_API_URL`. However, I think actually making the testdata file available to the container mimics better what happens in real life.

This is an alternative to #1582.


spencerschrock (approved):
Thank you!

PR #1594, opened 2025-10-20
sign with sigstore-go instead of cosign

spencerschrock:
This is in anticipation of Rekor v2, which will cause problems with the
current way we verify results in the webapp. In order to verify with
sigstore-go in the webapp,  we also need the bundle, so send that over
to the webapp.

Configuration was based on sigstore-go's signing example:
    
https://github.com/sigstore/sigstore-go/blob/b48a7c15af434de768f6db5d729a9aadafab5060/examples/sigstore-go-signing/main.go

As an added benefit, this drops binary size from 111MiB to 69MiB


signing/signing.go
```diff
@@ -202,15 +198,76 @@ func postResults(endpoint *url.URL, payload []byte) error {
 	return nil
 }
 
-func extractTlogIndex(bundlePath string) (int64, error) {
-	contents, err := os.ReadFile(bundlePath)
+func extractTlogIndex(bundle *protobundle.Bundle) (int64, error) {
+	// what does multiple tlog entries mean?
```
cmurphy: You would have multiple entries if you submitted the record to multiple logs, for instance if `root.SelectServices()` for rekor returned multiple logs from the signing config.

cmurphy (approved):
This looks good to me from a sigstore-go perspective, it matches with the example usage. Glad to see sigstore-go being used instead of cosign for this use case.

cmurphy (approved):

carloscanutocosta (approved):

carloscanutocosta (approved):

carloscanutocosta (approved):

